PR #10, opened 2021-08-12
Tooling: Provisioning scripts and docs for local development

### LINKED ISSUE #2: Tooling: Import demo content into the local dev environment
To assist with replicating the new design, we should provision the local dev environment with some actual posts from the wp.org News site.

coreymckrill:
This expands the provisioning scripts a bit to include additional
plugins used by the News site, configuring some WP settings, and
importing demo content. The content comes from the actual News site, but
it has been scrubbed of PII and there is a readme describing how to do
the scrub again if we want to update the demo content in the future.

It also adds more details in the main repo readme about how to set up
the dev environment.

To test, follow the instructions in the main readme file!

Fixes #2


readme.md
```diff
@@ -1,3 +1,29 @@
...
+* Node/npm
+* Yarn
+* Composer
+
+### Setup
+
+1. Set up repo dependencies: `yarn run initial-setup`
+2. Start up and provision the environment: `yarn run env:setup`
+3. Build the assets: `yarn workspaces run build`
+
```
iandunn, line 21:
```suggestion
1. Visit site at `localhost:8888`
1. Log in with username `admin` and password `password`

```

iandunn: I dunno why GH is removing the blank line and adding a new one, but I played around w/ [multi-line suggestions](https://github.blog/changelog/2020-02-26-multi-line-code-suggestions-beta/) and couldn't get it to not do that 🤷🏻 

Using `1.` in markdown will make the list automatically generate the numbers, instead of having to manually update them when things are added/removed from the list.

package.json
```diff
@@ -12,7 +12,13 @@
...
 		"initial-setup": "yarn && composer install && yarn run update-configs",
 		"update-configs": "TEXTDOMAIN=wporg composer exec update-configs",
-		"wp-env": "wp-env"
+		"wp-env": "wp-env",
+		"env": "wp-env start",
+		"env:stop": "wp-env stop",
+		"env:setup": "wp-env start --update && yarn run env:reset && yarn run env:import && yarn run env:misc",
+		"env:reset": "wp-env clean all && wp-env run cli \"wp site empty --yes\"",
+		"env:import": "wp-env run cli \"bash env/import.sh\"",
+		"env:misc": "wp-env run cli \"bash env/misc.sh\""
```
iandunn, line 22:
```suggestion
		"env:misc": "wp-env run cli \"bash env/misc.sh\""
		"env:cli": "wp-env run cli
```

Without this you have to SSH into the container to run arbitrary wp-cli commands

Adding that upstream to `repo-tools` would be nice too. That goes for some of the other suggestions too. I wonder if there's a more DRY & automation-friendly way than forking `repo-tools/templates/*/` 🤔 

Can `scripts` `extend` the scripts from repo-tools, or something like that? Not worth a lot of time right now, but maybe worth an issue to iterate later?

Maybe `/README.md` can just link to `/vendor/wporg/wporg-repo-tools/readme.md` for the common stuff? Or probably the github.com version, so they can see it before installing. `/env/README.md` could do something similar.

.wp-env.json
```diff
@@ -4,14 +4,22 @@
 		"SCRIPT_DEBUG": true,
 		"WP_DEBUG_LOG": "/tmp/wp-errors.log",
 		"FS_METHOD": "direct",
-		"WP_ENVIRONMENT_TYPE": "local"
+		"WP_ENVIRONMENT_TYPE": "local",
+		"JETPACK_DEV_DEBUG": true,
+		"WPORG_SANDBOXED": true
```
iandunn, line 9: Enabling auto-updates for all the non-VCS plugins would be convenient, so we don't have to manually run `composer update` frequently

Just a 'nice to have' though

.wp-env.json
```diff
@@ -4,14 +4,22 @@
...
 	},
 	"core": "WordPress/WordPress#5.8",
```
iandunn, line 11: Shouldn't this be `trunk`?

iandunn (approved):
LGTM, these are all minor comments

.wp-env.json
coreymckrill: Wouldn't having things get updated through WP screw up composer though? Like, the files and version wouldn't match the composer.lock file anymore?

.wp-env.json
coreymckrill: Yeah, it should be, but there's currently a bug in Gutenberg that causes an annoying notice in trunk. I'll open an issue to remind us to update this once it's fixed.

coreymckrill: https://github.com/WordPress/wporg-news-2021/issues/12

.wp-env.json
iandunn: What tangible problem would that create? I'd assume it'd only complain if you were running an `update` command, but doing that would make them match up again anyway, right?

.wp-env.json
iandunn: 👍🏻

package.json
coreymckrill: The suggested change was missing a comma so I added the same line in 77c5887

coreymckrill: Not sure about the upstream stuff, but could open an issue in `wporg-repo-tools` to explore the possibilities. It also might make sense to wait until we have multiple project repos depending on it before we add a bunch more stuff to it, so we have a better sense of what's needed.

iandunn: 👍🏻 , i'll open an issue there to discuss

.wp-env.json
coreymckrill: I just tried setting Gutenberg to an older version in the composer.json file and then updating it through the WP Admin UI and it broke. Not sure if that's a composer issue or a Docker mapping issue, but I think I'd prefer to leave it to `composer update` for now.

iandunn: Opened https://github.com/WordPress/wporg-repo-tools/issues/4 to discuss auto-updates, not a blocker here

Issue #6, opened 2021-08-12
Implement global header/footer design for News site


iandunn:
https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=4659%3A8203

It may be easiest to get it working for News first, and the roll it out to the rest of the network in #7. 

It should take into account any foreseeable needs for the network-wide rollout, though.

- [x] basic styles - mobile
- [x] basic styles - desktop
- [x] header menu
- [x] header search
- [x] overflow menu
- [x] responsive breakpoints
- [x] footer - mobile
- [x] footer - desktop

iandunn:
I was thinking of creating a "global header" dynamic block, and include that in `block-template-parts/header.html` (see https://github.com/WordPress/wporg-news-2021/issues/4#issuecomment-901306142). 

I don't think that's actually necessary, though. That block would contain the entire header area, and I'm assuming we're not going to want users changing the global header/footer. 

So, I think we can just skip FSE for the header/footer area, and treat it like a classic theme. The sidebar/content area would still be FSE, though.

How does that sound @coreymckrill ?

coreymckrill:
If we can get that to work, that's probably fine for now. I still don't have a great understanding at this point of how block templates are loaded, and if you can even mix and match php and html template parts...

iandunn:
Oh, huh, it looks like the `themes/wporg-news-2021/header.php` is never used; maybe Blockbase included it as a fallback or something?

If it's not used, then I think we'll need a header block after all. I'll take a closer look tomorrow.

coreymckrill:
I think it's only used in the 404.php template.

iandunn:
huh, but there's a `404.html` too. 

maybe it's just for translation? https://github.com/Automattic/themes/commit/ae7e63d133020db5df58584c1d019315dc2bc3a3

coreymckrill:
Ah, interesting. Good find.

coreymckrill:
> I was thinking of creating a "global header" dynamic block, and include that in block-template-parts/header.html

Thinking more about this. It does seem like this would be feasible, for both the header and the footer, it could just use a server side render and not have any options or UI. Maybe it would go in a separate plugin, `wporg-site-blocks` or something?

> I'm assuming we're not going to want users changing the global header/footer.

[...]

PR #20, opened 2021-08-20
Include global header block

iandunn:
this is a rough sketch of how the theme could include the global header from `repo-tools`

interconnected with https://github.com/WordPress/wporg-repo-tools/pull/9

see https://github.com/WordPress/wporg-news-2021/issues/6

if this sounds good, i'll add the rest


source/wp-content/themes/wporg-news-2021/functions.php
```diff
@@ -9,6 +9,13 @@
 add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_assets' );
 add_action( 'admin_init', __NAMESPACE__ . '\editor_styles' );
 
+
+define( 'REPO_TOOLS_DIR', dirname( ABSPATH ) . '/vendor/wporg/wporg-repo-tools' );
```
iandunn: this might need a different path for `wp-env` vs my local nginx env

production will definitely need a different path, since we'll have 1 checkout of `repo-tools`, instead of each theme having their own

iandunn:
refactored to use an mu-plugin approach, see https://github.com/WordPress/wporg-repo-tools/pull/9

interconnected w/ https://github.com/WordPress/wporg-mu-plugins/pull/4

cc @coreymckrill

source/wp-content/themes/wporg-news-2021/functions.php
```diff
@@ -2,6 +2,8 @@
 
 namespace WordPressdotorg\Theme\News;
 
+require_once WPORG_GIT_MUPLUGINS_DIR . '/mu-plugins/blocks/global-header-footer/blocks.php';
```
coreymckrill: From the theme's perspective, do you think we could just assume that the blocks are loaded and available? (And thus put this require statement in the sandbox file instead?) That's an assumption we already make a lot in other themes and plugins in wporg and wordcamp...

iandunn: 🤔 , I don't think we'd want to do that for everything in this repo, because some plugins will only be relevant to some sites.

We could do that for these particular blocks, though, since they intended to (eventually) be used on all sites. I think it'd be ideal to wait until most of the sites have been migrated to the new design, but don't feel strongly.

coreymckrill: I don't feel strongly either, I was just thinking that we could avoid adding an unnecessary line to the theme's functions.php file. But since we're using `require_once` it shouldn't throw any errors if the file is already loaded, so it's pretty harmless to leave it. 🤷

iandunn: Actually, now that I'm looking at how to make compatible across my local env, wp-env, and prod, I think it's just gonna be more practical to have `0-sandbox.php` load it 👍🏻 

The blocks don't show up in the Inserter -- presumably because of the missing `register_block_type()` arguments -- so we don't have to worry about users seeing it on non-FSE sites. 

And the performance impact isn't worth worrying about.

PR #21, opened 2021-08-24
Single post view

### LINKED ISSUE #17: Implement design for single posts
single-post

From https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=4659%3A8683

coreymckrill:
Adds block templates and template parts, tweaks the theme.json file, and adds some rules to the stylesheet to create the single post view as depicted in the mockup in #17.

This includes a responsive layout for the post title/meta area and the main post content column. The post content also accommodates wide and full alignment on individual blocks.

This doesn't include the global header or footer (see #6). It also doesn't include the social sharing links or like button shown in the mockup, because those elements are not yet available as site blocks. (There's [an issue](https://github.com/Automattic/jetpack/issues/19659) for the Like button, couldn't find one for social sharing icons.)

The mockup doesn't include a comment section, and it appears that the News site does not currently have comments enabled, so that has also been omitted.

Fixes #17 

## To test

1. The post used in the mockup is also included in the test content that gets added to the local dev environment for this repo, so you can easily compare the two.
2. Try changing the alignment of some blocks in the post to Wide and Full to see how those work.
3. Try changing the viewport size to find the responsive breakpoints.


source/wp-content/themes/wporg-news-2021/block-template-parts/content-single.html
```diff
@@ -0,0 +1,27 @@
+<!-- wp:group {"tagName":"article","layout":{"inherit":true}} -->
+<article class="wp-block-group">
+	<!-- wp:group {"tagName":"header","align":"wide","className":"entry-header"} -->
+	<header class="wp-block-group alignwide entry-header">
```
iandunn, line 4: 🤔 , is this a template that users will be able to edit?

If not, should we still use `group` blocks, etc, instead of just having `div`s?

It seems like we'd only need "content" blocks like `core/audio`, and functional ones like `core/query`, but the markup could just be regular HTML. That would cut down on the "comment soup" and make things more human readable.

source/wp-content/themes/wporg-news-2021/sass/ponyfill.scss
```diff
@@ -20,6 +20,8 @@
 @import "blocks/paragraph";
```
iandunn, line 20: Not directly related to this PR, but what's the reason to name this `ponyfill`? Was that a Blockbase artifact?

It seems like it's just the SASS index file; is it trying to say that it's ponyfilling all the missing CSS features like mixins etc?

It's kind of confusing, I think it'd be more intuitive to name it `style.scss`, `style.css`, `wporg-news-2021-style`, etc.

I can rename it after this PR is merged, if that sounds good to you.

source/wp-content/themes/wporg-news-2021/sass/ponyfill.scss
```diff
@@ -29,4 +31,6 @@
 @import "blocks/table";
 @import "blocks/video";
 @import "blocks/columns";
-@import "post/meta";
\ No newline at end of file
+@import "post/header";
```
iandunn, line 34: Huh, it looks like this when `SCRIPT_DEBUG` is `false`, but correct when it's `true`:

Screen Shot 2021-08-26 at 9 26 20 AM

🤔

source/wp-content/themes/wporg-news-2021/sass/blocks/_post-content.scss
```diff
@@ -0,0 +1,17 @@
+@include break-medium {
+	.single .wp-block-post-content > * {
```
iandunn, line 7:
```suggestion
	body.single .wp-block-post-content > * {
```

otherwise it's not clear which `.single` it's referring to

source/wp-content/themes/wporg-news-2021/sass/blocks/_post-content.scss
```diff
@@ -0,0 +1,17 @@
...
+		margin-left: calc(
+			min(
+				100% - var(--wp--custom--layout--content-size),
+				50% - (var(--wp--custom--layout--wide-size) / 2) + (var(--wp--custom--layout--wide-size) - var(--wp--custom--layout--content-size))
+			)
+		) !important;
```
iandunn, line 15: Is there a simpler way to do this? It doesn't feel like it should need to be the this complicated.

would grid be a better fit for all the layout? my understanding is that flexbox isn't really designed for that.

iandunn (changes requested):
The Grid and `SCRIPT_DEBUG` comments seem important to discuss more before merging, but I don't feel strongly about the others.

source/wp-content/themes/wporg-news-2021/block-template-parts/content-single.html
coreymckrill: I don't anticipate that this will be editable now. But I think it would be a mistake to make it incompatible (or at least less compatible) with FSE if that were to change in the future. I honestly don't think the legibility is impacted that much.

source/wp-content/themes/wporg-news-2021/sass/ponyfill.scss
coreymckrill: Yeah, that's an artifact from Blockbase. I'm fine renaming it, I agree it's kind of weird and confusing. 👍

source/wp-content/themes/wporg-news-2021/sass/ponyfill.scss
coreymckrill: That's odd, I'll see if I can replicate that...

source/wp-content/themes/wporg-news-2021/sass/ponyfill.scss
iandunn: 👍🏻 , i'll do that after this is merged.

source/wp-content/themes/wporg-news-2021/sass/ponyfill.scss
iandunn: huh, as long as it doesn't happen in production then I wouldn't worry about it 🤷🏻‍♂️

iandunn: Oh, I misread that, you said you were going to try to replicate, not that you'd tried and couldn't :)

coreymckrill: I can't reproduce this. I set `SCRIPT_DEBUG` to `false` in .wp-env.json and then confirmed that the constant was set to false on the front end, but the styles were still loading correctly...

source/wp-content/themes/wporg-news-2021/sass/blocks/_post-content.scss
coreymckrill: 🤔 Where else is `single` used in a different context?

iandunn: I can't think of one off the top of my head, but it seems common enough that a plugin etc could use it.

The main benefit IMO is the clarity, so someone isn't left asking "where does that come from? i don't see it anywhere in the theme...", and then have to dig around before they remember the body classes.

Not a big deal though

source/wp-content/themes/wporg-news-2021/sass/ponyfill.scss
iandunn: 👍🏻 , probably some quirk in my env then

source/wp-content/themes/wporg-news-2021/sass/blocks/_post-content.scss
coreymckrill: I couldn't think of a simpler way to ensure that the content column is responsively aligned on the right side, instead of centered, while still allowing individual blocks with `wide` or `full` alignment to expand appropriately.

This isn't actually using flexbox, just margins (same way Blockbase aligns content). I considered using flexbox, but it breaks things like having a left-aligned image with paragraph text wrapping around it. Plus it doesn't play well with the `.wp-container-[hash]` styles that get output automatically at the end of the document based on the `layout.contentSize` value in theme.json.

I haven't used CSS Grid much, but my understanding is that it doesn't work that well when you have an unknown number of elements in the grid. Plus I suspect it would have the same issues as flexbox regarding breaking some layouts...

iandunn: > This isn't actually using flexbox, just margins

I meant that `.entry-header`, `.entry-content` and `.post-navigation` are all creating identical columns, and the "like/share" section will too. This section is using margins to do it, but the others use Flex.

If we setup a Grid, though, they'd all automatically inherit the columns, rather than each one having to define them explicitly using various (and sometimes complication) methods.

> still allowing individual blocks with wide or full alignment to expand appropriately.

🤔 , I would think that should still work, but I can play around w/ it tomorrow.

> [grid might] breaks things like having a left-aligned image with paragraph text wrapping around it

I don't think that'd happen, since the grid styles wouldn't be inherited by the contents inside each cell

> [grid might not] play well with the `.wp-container-[hash]` styles 

🤔 , I see those in the source, but don't see them actually applying to anything. I'll see what happens when I play around w/ it tomorrow.


> it doesn't work that well when you have an unknown number of elements in the grid

I was picturing we'd use it to define the layout at a higher level, rather than the individual paragraphs. i.e.,

<table>
<tr> <td colspan="2">header</td> </tr>
<tr> <td colspan="2">local nav</td> </tr>
<tr> <td>entry meta</td> <td>entry title</td> </tr>
<tr> <td></td> <td>entry content</td> </tr>
<tr> <td>share</td> <td>like</td> </tr>
<tr> <td>previous</td> <td>next</td> </tr>
<tr> <td colspan="2">footer</td> </tr>
</table>

So we'd have a defined # of elements.

iandunn: I played around w/ it for a couple hours, but the big problem I ran into is that we really need Subgrid to make the (great-)grandchild elements of the grid container conform to the grid areas. [It's still not well supported](https://caniuse.com/?search=subgrid), though.

[`display:contents` could work](https://stackoverflow.com/questions/48050939/what-is-an-alternative-to-css-subgrid), but it's [hacky](https://rachelandrew.co.uk/archives/2017/07/20/why-display-contents-is-not-css-grid-layout-subgrid/) and also [not well enough supported](https://caniuse.com/css-display-contents).

IMO there are some excessive containers that could be removed, but many are semantically appropriate, and removing those would also be hacky.

Nested grids might work, but I don't think there's a significant enough benefit to make the time investment, since you've already got the flex/margin approach working.

So yeah, this seems good enough 👍🏻 

Some comments on the complex lines like this would be really appreciated, though -- both the _what_, and the _why_.

iandunn (approved):
The comments would be great, but not a blocker. Thanks!

source/wp-content/themes/wporg-news-2021/sass/blocks/_post-content.scss
coreymckrill: Good idea. Comments added in 0488417

iandunn: Thanks!

PR #28, opened 2021-09-07
Refine padding/margin, override block gap

iandunn:
See https://github.com/WordPress/wporg-news-2021/issues/6
Interconnected with https://github.com/WordPress/wporg-mu-plugins/pull/5


source/wp-content/themes/wporg-news-2021/block-template-parts/post-navigation.html
```diff
@@ -2,13 +2,14 @@
 <hr class="wp-block-separator alignfull has-text-color has-background has-light-grey-background-color has-light-grey-color is-style-wide"/>
 <!-- /wp:separator -->
 
-<!-- wp:group {"tagName":"nav","align":"wide","className":"post-navigation"} -->
-<nav class="wp-block-group alignwide post-navigation">
+<!-- wp:group {"tagName":"nav","align":"full","className":"post-navigation"} -->
+<nav class="wp-block-group alignfull post-navigation">
```
coreymckrill, line 6: Why change this to full? Won't this make it so the previous/next links are out on the edges of the viewport instead of staying within the content column?

iandunn: 🤔 , good point. IIRC I did it so that the content would line up w/ the footer content, but i'll take a closer look.

iandunn: yeah, i think this works fine. we could also change it to work w/ `alignwide`, but i think this makes more sense semantically, since post navigation isn't part of the content of a post 

_(at least it's not in the DOM; philosophically i can see it either way, but would personally learn towards it being outside)_

iandunn:
merged `trunk` after #31, looks clean

source/wp-content/themes/wporg-news-2021/block-template-parts/post-navigation.html
iandunn: I went ahead and merged to keep the # active branches low, but I'm happy to continue discussing if you have any other thoughts

coreymckrill: Just checked this on a super wide screen, and it looks like it does make the previous/next links sit too far out. My read on the design mockup is that the "previous" link should line up with the post date and author on the left side, and the "next" link should be justified right and line up with the edge of the post title and content column on the right side.

iandunn: Ah, I see what you're saying.

I wonder if it should be aligned with the post content _and_ the header/footer, and that the real problem here is that there isn't a "max width" of sorts on the header/footer. I'm guessing we don't want those to grow endlessly.

@beafialho can you add a Figma example of what a whole page should look like when the browser is `2000px` wide?

beafialho: Sure, here's what it would look like in a 2000px browser. The behaviour should be: post navigation aligns with the edges of the grid. The header is not included in this grid, but the footer is. Does it make sense?

Captura de ecrã 2021-09-30, às 16 37 24

Captura de ecrã 2021-09-30, às 16 37 03



[Prototype](https://www.figma.com/proto/O9I8Jl7kaUMTX39LlRGbGO/News-Page?page-id=4659%3A8682&node-id=7987%3A32037&scaling=scale-down-width&starting-point-node-id=7987%3A32037)
[Mockup](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=7987%3A32037)

iandunn: thanks!

moved discussion to https://github.com/WordPress/wporg-news-2021/issues/37 since this pr is already merged

Issue #29, opened 2021-09-08
Gutenberg 11.4 broke the single post layout


coreymckrill:
On Gutenberg 11.3 (when I built the single post layout), the post header looked like this:
![before](https://user-images.githubusercontent.com/916023/132594301-89faf1e6-d10a-4253-bd07-7aae5e149f0c.jpg)

Now that 11.4 is out, it looks like this, and doesn't align correctly on some screen widths:
![after](https://user-images.githubusercontent.com/916023/132594320-af3fce01-f20c-4283-b3c6-91165ed173db.jpg)

The difference seems to be that the "header" group block is now receiving the `wp-container-[hash]` class, even though it does not have `"layout":{"inherit":true}` in its block properties. The parent group block does have this property.

I'm still trying to track down the change in Gutenberg that caused this to determine if it's a regression or a feature...

coreymckrill:
Looks like this is what broke it: https://github.com/WordPress/gutenberg/pull/34194

coreymckrill:
Fixed by #30

Issue #25, opened 2021-09-01
Audit use of `!important`


iandunn:
We have [a lot of instances where `!important` is used](https://github.com/WordPress/wporg-news-2021/search?q=%22%21important%22), and I've already run into a case where that's causing a problem.

Most of them appear to be from https://github.com/WordPress/wporg-news-2021/commit/05ec95d16b4c2395d5680168785153cf4f8ff876, so those may not actually be necessary for us.

The only other I see is from https://github.com/WordPress/wporg-news-2021/commit/7241d17afdf33e030fd5ce58bb88ceac5784450b, and I'm guessing there's a good reason for it - cc @coreymckrill

coreymckrill:
The usages in 7241d17 are necessary to override some `!important` rules that are generated and output by Gutenberg (or maybe Core?) based on the `contentSize` and `wideSize` values set in theme.json.

I'm guessing the ones that came from Blockbase are there for a similar reason...

iandunn:
I haven't noticed any side effects from 2d83fbf, so i think in some cases the blockbase ones aren't relevant to our custom design. in general, i suspect there may be a lot of blockbase styles that go unused. IIRC Lighthouse will identify rules that aren't used on any page, so we could run that before the launch.

in some cases i've be able to override the `container-{hash}` styles by using a more specific selector instead of `!important`, or by removing `layout: inherit` from the block comment. not sure if that'd work w/ 7241d17 though.

tellyworth:
@kellychoffman @beafialho any recommendations here?

kellychoffman:
Not sure the best approach here - will lean on theme devs for help with this.

iandunn:
Lately I've been using selectors like `.query-navigation-container[class*="wp-container-"]` to gain a higher specificity than Gutenberg's layout styles, and that's been working well.

That may still be a symptom of a deeper problem, though, like hardcoding styles instead of taking advantage of things Gutenberg could automatically if we knew how to leverage that.

xref https://github.com/WordPress/wporg-news-2021/pull/30#discussion_r705556322, https://github.com/WordPress/wporg-news-2021/pull/30#discussion_r705625816

PR #30, opened 2021-09-09
Fixes for breaking changes from Gutenberg 11.4

coreymckrill:
Our theme had some alignment styles from Blockbase that didn't play well with changes introduced in Gutenberg 11.4. I went and found the commit where Blockbase was fixed to accommodate those changes and brought it over, and it seems to mostly fix the problem described in #29.

This is still a bit of a work-in-progress as I try to figure out if I can avoid introducing more `!important` clauses to the stylesheet, but I wanted to get your eyes on this @iandunn because it also appears to affect the global header (currently it's getting some margin added on the sides instead of being full-width).

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/blocks/_post-title.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/blocks/_post-title.scss
@@ -1,3 +1,3 @@
 .wp-block-post-title {
-	margin-top: 0;
+	margin-top: 0 !important; // TODO this might be fixable via the blockGap prop in theme.json.
 }
--- a/source/wp-content/themes/wporg-news-2021/sass/blocks/_post-title.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/blocks/_post-title.scss
@@ -1,3 +1,5 @@
 .wp-block-post-title {
-	margin-top: 0 !important; // TODO this might be fixable via the blockGap prop in theme.json.
+	.single [class*="wp-container-"] & {
+		margin-top: 0;
+	}
 }
--- a/source/wp-content/themes/wporg-news-2021/sass/post/_header.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/post/_header.scss
@@ -6,12 +6,14 @@
 	.wp-block-post-author,
 	.wp-block-post-terms {
 		display: inline-block;
+		margin-top: 0 !important; // TODO this might be fixable via the blockGap prop in theme.json.
 	}
 }
 
 @include break-large {
 	.entry-header {
 		display: flex;
+		align-items: flex-start !important; // TODO eventually this might be a layout setting.
 		justify-content: space-between;
 
 		.entry-meta {
--- a/source/wp-content/themes/wporg-news-2021/sass/post/_header.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/post/_header.scss
@@ -6,7 +6,10 @@
 	.wp-block-post-author,
 	.wp-block-post-terms {
 		display: inline-block;
-		margin-top: 0 !important; // TODO this might be fixable via the blockGap prop in theme.json.
+
+		[class*="wp-container-"] & {
+			margin-top: 0;
+		}
 	}
 }
 
--- a/source/wp-content/themes/wporg-news-2021/block-template-parts/content-single.html
+++ b/source/wp-content/themes/wporg-news-2021/block-template-parts/content-single.html
@@ -1,6 +1,6 @@
 <!-- wp:group {"tagName":"article","layout":{"inherit":true}} -->
 <article class="wp-block-group">
-	<!-- wp:group {"tagName":"header","align":"wide","className":"entry-header"} -->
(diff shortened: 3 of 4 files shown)


iandunn (approved):
LGTM 👍🏻 

great job on the docs, and thanks for the early heads up 👍🏻

source/wp-content/themes/wporg-news-2021/sass/blocks/_post-title.scss
```diff
@@ -1,3 +1,3 @@
 .wp-block-post-title {
-	margin-top: 0;
+	margin-top: 0 !important; // TODO this might be fixable via the blockGap prop in theme.json.
```
iandunn: yeah, i wonder if we should just set it to `0`? pretty much every element in the header and footer will need it manually turned off. 

or maybe we have a single css rule that disables it for the header, footer, and all their descendents? it could be useful for the content area, just not in the locked templates

https://github.com/WordPress/gutenberg/pull/33812

source/wp-content/themes/wporg-news-2021/sass/post/_header.scss
```diff
@@ -6,12 +6,14 @@
 	.wp-block-post-author,
 	.wp-block-post-terms {
 		display: inline-block;
+		margin-top: 0 !important; // TODO this might be fixable via the blockGap prop in theme.json.
```
iandunn: not sure if it'd work here, but the header didn't need the `!important`, just:

```css
.wp-block-group.site-header {
	margin-top: 0;
}
```

source/wp-content/themes/wporg-news-2021/sass/blocks/_post-title.scss
iandunn: this is interesting:

https://every-layout.dev/layouts/stack/

iandunn: > a single css rule

this seems to work so far:

```css
[class*="wp-container-"].site-header,
[class*="wp-container-"].site-footer,
[class*="wp-container-"].site-header > * + *,
[class*="wp-container-"].site-footer > * + *,
.site-header [class*="wp-container-"] > * + *,
.site-footer [class*="wp-container-"] > * + * {
    margin-top: 0;
}
```

...but now i have to update _every_ individual element inside the header/footer that is supposed to have a margin other than `0`, and make the selectors more specific. maybe there's a better way?

source/wp-content/themes/wporg-news-2021/sass/post/_header.scss
iandunn: i think this might work universally

```css
[class*="wp-container-"] .site-footer__logos-container {
    margin-top: 0;
}
```

xref #25

coreymckrill: What do you mean universally?

iandunn: sorry, i meant we might always be able to do 

```css
[class*="wp-container-"] .{ whatever element you want to target } {
    margin-top: 0;
}
```

so instead of the `!important` rules we'd have

```css
[class*="wp-container-"] .wp-block-post-author,
[class*="wp-container-"] .wp-block-post-terms,
[class*="wp-container-"] .wp-block-post-title {
    margin-top: 0;
}
```

iandunn: in some cases it might be better to override the gap for specific blocks in `theme.json`, though?

xref https://github.com/WordPress/gutenberg/pull/33991

source/wp-content/themes/wporg-news-2021/sass/blocks/_post-title.scss
iandunn: As we run into problems with this, it'd be good to add them here:

https://github.com/WordPress/gutenberg/issues/34716

source/wp-content/themes/wporg-news-2021/sass/post/_header.scss
coreymckrill: Ah, I see. This definitely sounds better than using `!important`.

> in some cases it might be better to override the gap for specific blocks in theme.json, though?

I would expect it would only make sense to use theme.json if you _always_ want the gap to be different, regardless of what context the block is in.

source/wp-content/themes/wporg-news-2021/sass/blocks/_post-title.scss
coreymckrill: Removed the `!important` here in 546e280

source/wp-content/themes/wporg-news-2021/sass/post/_header.scss
coreymckrill: Removed the `!important` here in 546e280

coreymckrill:
@iandunn Do you want to address the global header margins in the PR, or in a separate one? Here's what it's looking like right now:

global-header

iandunn:
I'll take care of it in #28 , I've got it mostly working locally, but am playing around with a few approaches

PR #33, opened 2021-09-16
Add Browsersync to improve devex

### LINKED ISSUE #10: Enable live reload and hot module reload
IIRC wp-scripts has had live reload for awhile, and recently added HMR

iandunn:
it's a pain to manually refresh, and slows me down

i tried to use wp-scripts first, but it kind of assumes you're working on a block where css is `@imported` from a jsx file, and working w/ multiple mu-plugins didn't seem practical 

mostly fixes https://github.com/WordPress/wporg-repo-tools/issues/10 - it's annoying that it's an extra watch script to run, rather than being included in the yarn workspace. i couldn't get that working though, and it doesn't seem worth it to go down that rabbit hole right now

### DIFF
--- a/bs-config.js
+++ b/bs-config.js
@@ -1,0 +1,36 @@
+// See http://www.browsersync.io/docs/options/
+
+const paths = {
+	theme: 'source/wp-content/themes/wporg-news-2021',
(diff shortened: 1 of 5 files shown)


iandunn:
@coreymckrill any thoughts on the general direction of this?

coreymckrill (approved):
Seems like a decent solution for now 👍

PR #41, opened 2021-10-02
Add the Blue Bar

### LINKED ISSUE #32: The Blue Bar
The Blue Bar is the "local" navigation bar that appears right below the global header. On most views it just looks like this:

blue-bar-1

Breadcrumb on the left, category dropdown on the right.

But sometimes it's expanded so that some or all of the breadcrumb is emphasized in a larger font:

blue-bar-all-posts

blue-bar-podcast

And then, for some category archives, it's a different color:

blue-bar-red

blue-bar-green

blue-bar-charcoal

blue-bar-white

The consistent things are:

* Breadcrumb
* Category dropdown
* The paint stroke on the bottom edge (should be able to make this an SVG and modify its color with CSS)

Does it make sense to try and build this bar as a template part, and contain the logic for the variations within it? Or would it be better to construct the bar in each page-level template separately? (i.e. single.html, page.html, archive.html, etc. all have their own markup for the bar)

coreymckrill:
Introduces the "Blue Bar" or, more accurately, header navigation. It includes a simple taxonomy-based breadcrumb on the left and a dropdown menu of categories on the right. The color scheme changes ("The Not Blue Bar") for certain categories. The layout responsively handles mobile, tablet, and desktop views.

The categories dropdown functionality is entirely CSS-based. I'm not sure if this is the right approach, as it made the stylesheet a bit more complex, but it's nice that no JS file is required.

This also makes the entire site header (including the global header) fixed to the top when not in a narrow mobile viewport, as called for in the design.

This does not yet handle the "All Posts" large heading for the Posts Page view, due to the issue described below.

Fixes #32


coreymckrill:
Noting that there currently isn't a very good way to show the "All Posts" title, which should appear in the Blue Bar on the Posts page if the site is configured to have a static homepage. Ideally we'd use the `query-title` block since that outputs the category term names when viewing category archives. I suggested that this block should also cover the Posts page title [here](https://github.com/WordPress/gutenberg/issues/33476#issuecomment-943811086).

The alternative for now, I think, is to copy the `header-navigation.html` block template part directly into a template specifically for the Posts page view (`home.html` I think?) and replace the `query-title` block with a hard-coded "All Posts" heading block.

source/wp-content/themes/wporg-news-2021/block-templates/index.html
```diff
@@ -1,12 +1,6 @@
...
-<!-- wp:group {"tagName":"header","className":"header-archive header-archive-all-posts","layout":{"inherit":true}} -->
-<header class="wp-block-group header-archive header-archive-all-posts">
-	<!-- wp:heading {"level":1,"align":"wide"} -->
-	<h1 class="alignwide">All Posts</h1>
-	<!-- /wp:heading -->
-</header>
-<!-- /wp:group -->
+<!-- wp:template-part {"slug":"header","align":"full","className":"site-header-container"} /-->
+<!-- This is a hack to ensure content isn't covered by the fixed header -->
+<!-- wp:template-part {"slug":"header","align":"full","className":"site-header-offset"} /-->
```
coreymckrill, line 3: @iandunn If there's a better way to do this, I'm certainly open to it. But if we do go with this approach, one thing that will need to be changed is how the template file for the global header is included. Right now it uses `require_once`, but that means that its markup doesn't get output a second time within `site-header-offset` because it's already been loaded.

coreymckrill:
> The alternative for now, I think, is to copy the header-navigation.html block template part directly into a template specifically for the Posts page view (home.html I think?) and replace the query-title block with a hard-coded "All Posts" heading block.

On second thought, the "All Posts" heading block could probably live outside the header.html block template part. So the home.html file would be identical to index.html, but with the addition of a heading block above the `<main>` tag. We'd have to figure [this](https://github.com/WordPress/wporg-news-2021/pull/41/files#r733194398) out first, though.

source/wp-content/themes/wporg-news-2021/block-template-parts/header-navigation.html
```diff
@@ -0,0 +1,27 @@
+<!-- wp:group {"align":"wide","className":"site-navigation"} -->
```
iandunn: `site-navigation` feels vague enough to be confused with the global header's `site-header__navigation`. should we rename them to something like `global-header__navigation` and `local-header__navigation`?

_( child elements of each would be renamed too )_

source/wp-content/themes/wporg-news-2021/block-template-parts/header-navigation.html
```diff
@@ -0,0 +1,27 @@
...
+			<!-- wp:home-link {"label":"News"} /-->
+		<!-- /wp:navigation -->
+		<!-- wp:query-title {"type":"archive","className":"site-navigation-breadcrumb-current"} /-->
+	</div>
+	<!-- /wp:group -->
+
+	<!-- wp:group {"className":"site-navigation-categories"} -->
+	<div class="wp-block-group site-navigation-categories">
+		<!-- wp:html -->
+		<input type="checkbox" id="site-navigation-categories-toggle" class="site-navigation-categories-toggle" aria-label="Open categories list" />
```
iandunn: Should this be a `button` instead? Or have a `form` if it's going to stay an `input`?

or use the same approach as global header menu, for consistency?


todo need to understand more

source/wp-content/themes/wporg-news-2021/block-templates/index.html
iandunn: todo test to understant why doing this

--

clarify docs why it's needed
open G issue for bug or whatever, link to it in comment

source/wp-content/themes/wporg-news-2021/sass/post/_header.scss
```diff
@@ -1,4 +1,240 @@
-.header-archive {
+.site-header {
```
iandunn: Could this be modularized into 2-4 files? It seems like a lot for 1 file, and might have different/unrelated chunks that could be separated.

Is `post/header/scss` the right file, it seems like this applies to pages/archives/etc? i'd assume this file would just have `entry-header` stuff

iandunn:
> made the stylesheet a bit more complex, but it's nice that no JS file is required.

i'll check when i get into the code, but I generally prefer that tradeoff too, seems less fragile and more straightforward. might be more reliable for screen readers too?

> makes the entire site header (including the global header) fixed to the top 

* should that style go in the block stylesheet, since it'll be needed when we roll the global header out?
* `<rant>`i hate fixed navigation, it's one of my biggest pet peeves on the web.`</rant>`. nothing actionable there, just needed to say it 😆 

>  copy the header-navigation.html block template part directly into a template specifically for the Posts page view 

Another idea would be to have both `query-title` and hardcoded `All Posts` in the template part, but hide one or the other w/ CSS until the G issue is resolved. No preference, just throwing it out there.

source/wp-content/themes/wporg-news-2021/block-template-parts/header-navigation.html
coreymckrill: I like those naming conventions 👍 

I'll update to `local-header` in this PR.

source/wp-content/themes/wporg-news-2021/block-template-parts/header-navigation.html
coreymckrill: > Should this be a button instead? Or have a form if it's going to stay an input?

The reason it's a checkbox is so that it can have a toggleable state that will persist when not in focus (unlike hovering over or focusing on the element). I'm not aware of a way to do that with a button.

As far as I can tell, an `input` element [does not](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#technical_summary) have to have a `form` element parent to be valid HTML.

> or use the same approach as global header menu, for consistency?

Possibly. I didn't think the navigation block would allow for the type of dropdown functionality that the design calls for here, but I could be wrong.

source/wp-content/themes/wporg-news-2021/sass/post/_header.scss
coreymckrill: Yeah, I've started to think this should all go in its own file(s). This could go along with the [idea](https://github.com/WordPress/wporg-news-2021/issues/45#issuecomment-944782370) of adding a `components` directory.

coreymckrill:
> should that style go in the block stylesheet, since it'll be needed when we roll the global header out?

I guess the question is if the global header will always be fixed, or if that's specific to the News site...

> Another idea would be to have both query-title and hardcoded All Posts in the template part, but hide one or the other w/ CSS until the G issue is resolved. No preference, just throwing it out there.

🤔 That might work. I'll try it.

source/wp-content/themes/wporg-news-2021/block-template-parts/header-navigation.html
iandunn: 👍🏻 , i'll update the global header in https://github.com/WordPress/wporg-mu-plugins/pull/15

source/wp-content/themes/wporg-news-2021/block-template-parts/header-navigation.html
iandunn: huh, this and the others were supposed to be drafted comments while i looked into it. maybe i accidentally published it, or gh published the drafted review when i published the comment in the main thread?

source/wp-content/themes/wporg-news-2021/sass/post/_header.scss
```diff
@@ -1,4 +1,240 @@
...
+	&-container {
+		width: 100%;
+
+		// Matches the breakpoint for the admin bar.
```
iandunn:
```suggestion
		// Don't fix the header on mobile, because screen real estate is limited.
		// Matches the breakpoint for the admin bar.
```

source/wp-content/themes/wporg-news-2021/block-templates/index.html
iandunn: This is a creative approach, but it seems safe to assume that the local nav will always have a known/stable height. If that's the case, then this seems to solve the problem in a more conventional and obvious way:

```css
.site-content-container {
   // Fixed position elements are removed from the normal document flow, so we have to account for it in order to avoid content being hidden under it.
   margin-top: var( --wp-site-header-height );
}
```

Is that not the case?

source/wp-content/themes/wporg-news-2021/sass/post/_header.scss
iandunn: 🤔  it seems like we might want both `posts/` and `components/`? Most of the stuff in `post/` right now seems legitimately tied to posts (and only posts).

Stuff like this isn't directly tied to posts, though

source/wp-content/themes/wporg-news-2021/sass/post/_header.scss
```diff
@@ -1,4 +1,240 @@
...
+			span {
+				padding: 0 3px 0 0;
+			}
+
+			svg {
+				vertical-align: middle;
+			}
+		}
+
+		&-toggle {
```
iandunn: The partial class names feels kind of awkward to me; if I don't remember what the containing SCSS block is named, then I have to scroll back up to the top to find out.

PhpStorm does show it in its breadcrumb toolbar, but it still feels less readable to me, especially in long blocks like this.

I don't feel strongly though.

source/wp-content/themes/wporg-news-2021/block-template-parts/header-navigation.html
iandunn: Ah, now that I understand this, I actually really like it 👍🏻 

It doesn't seem like a common approach, though, so comments here and/or in the CSS would probably be a kindness to our future selves & others.

source/wp-content/themes/wporg-news-2021/sass/post/_header.scss
```diff
@@ -1,4 +1,240 @@
...
+		}
+
+		@include break-xlarge {
+			padding: 0;
+		}
+	}
+
+	&-container {
+		--bar-background-image: url("assets/svg/brush-stroke-nav-bar-blue-1.svg"),
+			url("data:image/svg+xml,%3Csvg height='1000' width='1' version='1.1' xmlns='http://www.w3.org/2000/svg'%3E%3Crect x='0' y='0' height='1000' width='1' r='0' rx='0' ry='0' fill='%233e58e1'%3E%3C/rect%3E%3C/svg%3E");
```
iandunn: Creating a locally-scoped var for this is nice to help readability 👍🏻

iandunn (approved):
Left some comments, but no blockers 👍🏻

iandunn:
>  the question is if the global header will always be fixed, or if that's specific to the News site...

The former seems like a safe assumption. @beafialho, is that what you were thinking?

source/wp-content/themes/wporg-news-2021/block-templates/index.html
coreymckrill: The local nav should mostly have a stable height (expanded category list notwithstanding, but that's intended to overlap the content anyway). However, on narrow screens, the height changes when the breadcrumb and the categories toggle stack. Furthermore, the global header (also fixed) can also have a variable height at different viewport widths, if I'm not mistaken.

iandunn: Ah, you're right. I don't see any good CSS-only approaches for the dynamic height problem ☹️ 

I wouldn't be surprised if we could hardcode an educated guess that'd work in practical use cases, but the duplicate-content approach WFM too.

In that case, it might help to link to something like https://stackoverflow.com/questions/26163818/css-fixed-header-of-unknown-height/26164057 in the comments, just to clarify the problem and the tradeoffs with various solutions

beafialho:
Yes @iandunn I was thinking the global header would be fixed for all the site's pages.

source/wp-content/themes/wporg-news-2021/block-template-parts/header-navigation.html
iandunn: So, something like this?

```html
<div class="site-header"
    <header class="global-header"
        <nav class="global-header__navigation"
    <nav class="local-header"
        <div class="local-header__navigation"
```

coreymckrill: Done in ff43619

source/wp-content/themes/wporg-news-2021/sass/post/_header.scss
coreymckrill: Moved to `components` in ff43619

source/wp-content/themes/wporg-news-2021/block-template-parts/header-navigation.html
coreymckrill: Added some comments in fa4f849

source/wp-content/themes/wporg-news-2021/block-templates/index.html
coreymckrill: Added a link in fa4f849

coreymckrill:
I was just reading [this](https://ishadeed.com/article/less-absolute-positioning-modern-css/), and thinking that after all this, it might actually be simpler and more robust to use CSS grid for the layout of this bar and manage the positions of things responsively. It might not be worth the refactor time at this point, though.

@iandunn I'm going to go ahead and merge this, and you can revisit with CSS grid if layout bugs crop up later.

PR #49, opened 2021-10-21
Standardize left/right edge alignment

### LINKED ISSUE #37: Refine edge and content alignment
There are a couple related discussions in closed threads, so consolidating here.

* Edge alignment inconsistent - https://github.com/WordPress/wporg-news-2021/pull/28#discussion_r717030878
* Content column no longer aligned w/ post title - https://github.com/WordPress/wporg-news-2021/commit/fa25d31b2dbaf3bde87a3d90cb568e0505cd4c98#r57269081

### goals

- [ ] all top-level elements should align on their left/right edges, including local header. only exception is global header.
    * text in post content isn't justified, so the right edge won't appear to align, but the container does
    * body text extra padding to push towards center of viewport. characters per line on desktop can be as high as `121`
- [ ] global header elements align to edges of screen, with a small gap
- [ ] `alignwide`
    * quotes - same edge padding as top-level elements, text inside has larger left margin (not aligned w/ anything). characters per line on desktop about `48`
        * on mobile, the background image is offset so that half of the first mark of off hidden
- [ ] `alignfull`:
    * images and videos - go to viewport edge
    * normal text - background goes to viewport edge, text is aligned w/ other page elements
         * characters per line on desktop is as high as `116`
    * verse - background fills content area, text inside has similar padding to edge padding
- [ ] in general, mobile is same as above, just with smaller edge padding 

@beafialho, @coreymckrill, does all that seem right to you?

@beafialho, what are your thoughts on the characters per line? they seem much higher than the usual guideline

### implementation

- [ ] It seems to make the most sense to remove all left/right padding/margin from all top level components, and then set a single rule -- ie, `body > * {...}` -- to apply the new one. the header can overwrite it, or we could do `*:not(.site-header)` or whatever.
- [ ] also move left/right padding from `alignfull` to  only _text_ children (not images/videos/etc)
- [ ] and then make all the various changes to implement the details of the mockup.

any thoughts @coreymckrill ?

iandunn:
* `alignfull` layout elements like `.post-navigation` should stretch to the viewport edge (but with internal padding). Compare to the the revised mockup in https://github.com/WordPress/wporg-news-2021/pull/28#discussion_r719533873
* Post titles should align with the post body.
* alignfull content should fill the screen
* alignwide content should be larger than normal content, but still have left/right spacing

`footer-archive` still needs some work, but that can be fixed in the pixel-perfect milestone.

The page archive also needs some refinement (removing post-specific blocks, etc), but that can be done later.

Fixes #37
Interconnected with https://github.com/WordPress/wporg-mu-plugins/pull/15


source/wp-content/themes/wporg-news-2021/theme.json
```diff
@@ -338,24 +341,6 @@
...
 				}
 			},
-			"post-content": {
-				"padding": {
-					"top": "32px",
-					"bottom": "70px",
-					"left": "var(--wp--style--block-gap)",
-					"right": "var(--wp--style--block-gap)"
-				},
-				"breakpoint": {
```
iandunn, line 348: These were removed because they're no longer used.

If they were needed, though, I think the `edge-spacing` var this PR introduces is better than breakpoint-specific vars.

See https://github.com/WordPress/gutenberg/issues/27107#issuecomment-731151158

iandunn:
reminder: this will need to be rebased after #41 is merged

scruffian:
This seems problematic for the "People of WordPress" section:
Screenshot 2021-12-07 at 12 04 01

iandunn:
Can you elaborate on that?

scruffian:
Sorry I was misunderstanding the implementation, its fine :)

PR #51, opened 2021-10-22
Theme: Add template stubs.

iandunn:
This is just the markup, so we have a foundation to iterate on.

See #8

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,8 @@
 		"env:reset": "wp-env clean all && wp-env run cli \"wp site empty --yes\"",
 		"env:import": "wp-env run cli \"bash env/import.sh\"",
 		"env:misc": "wp-env run cli \"bash env/misc.sh\"",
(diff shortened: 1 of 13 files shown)


iandunn:
On the on hand, it feels helpful to have the templates and parts with rough markup, in order to have a sense of how everything fits together, where stuff goes, etc. 

For example, some categories can use the generic `category.html`, and just have some tweaks w/ CSS. Others categories need custom markup, though.

_( aside: that ends up w/ duplicate templates b/c of https://github.com/WordPress/gutenberg/issues/32939 )_

On the other hand, adding these too early might result in old markup being used; when we update it in the future, we'll have to update a lot of templates.

Overall it seems worth it to me, though. Any thoughts @coreymckrill ?

iandunn:
Doh, I thought your AFK started next week, rather than this one. No worries on reviewing this, I'm sure you've got a lot on your plate right now.

PR #54, opened 2021-10-26
"Listen All Episodes" is weird english

dd32:
I'm not sure how to best rephrase this, "Listen to All Episodes"? "View All Episodes"? "All Episodes"?

The existing text reads weird to me, but the link isn't "listen" or "view"  in the media term, rather it brings you to the list of podcast episodes.. so perhaps "All Episodes" is better here?

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/block-template-parts/footer-archive.html
+++ b/source/wp-content/themes/wporg-news-2021/block-template-parts/footer-archive.html
@@ -65,7 +65,7 @@
 			</ul>
 			<!-- /wp:social-links -->
 			<!-- wp:paragraph {"className":"podcast-link"} -->
-			<p class="podcast-link"><a href="https://wordpress.org/news/podcast/">View All Episodes</a></p>
+			<p class="podcast-link"><a href="https://wordpress.org/news/podcast/">Listen to all episodes</a></p>
 			<!-- /wp:paragraph -->
 		</div>
 		<!-- /wp:group -->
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-news-2021/block-template-parts/footer-archive.html
```diff
@@ -65,7 +65,7 @@ <h4>WP Briefing &mdash; The WordPress Podcast</h4>
 			</ul>
 			<!-- /wp:social-links -->
 			<!-- wp:paragraph {"className":"podcast-link"} -->
-			<p class="podcast-link"><a href="https://wordpress.org/news/podcast/">Listen All Episodes</a></p>
+			<p class="podcast-link"><a href="https://wordpress.org/news/podcast/">View All Episodes</a></p>
```
dd32:
```suggestion
			<p class="podcast-link"><a href="https://wordpress.org/news/podcast/">All Episodes</a></p>
```

iandunn:
Yeah, it's definitely awkward 👍🏻  

I like "listen" because it points out the distinction between text and audio content, but you're right that it's not descriptive of where the link actual takes you (at least not literally/directly). Maybe that's ok, though? I don't have strong feelings.

@beafialho , do you have any thoughts?

beafialho:
Thanks for flagging this.

It makes sense to me to go with "Listen to all episodes" because the ultimate intent is to listen to these post's content, which is audio, even if you find the post list first. It sounds weird IMO not having a described action, such as "Listen", "View" or "Go to", in a button.

Issue #80, opened 2021-11-17
Fix pagination number `found_posts`


iandunn:
After 483f34db1518a6f09f203f3c5f953de0c245c1e8, the `wp:query-pagination-numbers` block shows 10 pages when it should show 11.

I think the problem is the one described at https://codex.wordpress.org/Making_Custom_Queries_using_Offset_and_Pagination#Offset_Using_pre_get_posts, where `found_posts` needs to be adjusted to account for the new `offset`.

```php
add_filter( 'found_posts', __NAMESPACE__ . '\foo', 1, 2 );
function foo( $found_posts, $query ) {
	$is_posts_index = is_posts_index( $query );
	
	if ( ! $is_posts_index ) {
		return $found_posts;
	}
	
	return $found_posts + 5;
}
```

This is blocked by #70 , though.

coreymckrill:
#70 is closed now. Is this still blocked @iandunn?

iandunn:
No, but it looks like the problem was also fixed by some earlier change 👍🏻

Issue #76, opened 2021-11-17
community - final markup and css


tellyworth:

iandunn:
#88 did a first pass at the markup, but it sounds like there are more issues to resolve

MaggieCabrera:
I think https://github.com/WordPress/wporg-news-2021/pull/199 closes this one?

ryelle:
I agree - if there are any issues with the page, we should open new issues with details/screenshots.

Issue #70, opened 2021-11-05
Fix broken pagination links


iandunn:
It looks like all the uses of pagination are broken.

- [ ]  `front-page/` template parts - "See all ..." links to `/?query-page=2` instead of `/page/2`. That might be a bug with `wp:query-pagination-next`, or maybe we're using it wrong. See #55
- [x] archive/category templates - links to pages have broken `query-page` param instead of `page/n` format
- [x] archive/category templates - shown even when cat only has 1 post
- [ ] Look for those problems in other templates too, and fix any instances

jeffikus:
@scruffian @matiasbenedetto is this related to any of the work we've been doing on pagination? Anything we can use from there to help here?

scruffian:
No, this is unrelated

ryelle:
It looks like if you use `inherit` on the main posts query, the pagination link on "See All Posts" should link to /page/2/ https://github.com/WordPress/gutenberg/issues/32531#issuecomment-877992554

```html
<!-- wp:query {"query":{"perPage":5,"inherit":"true"},"tagName":"section","className":"front__latest-posts","align":"full"} -->
```

tellyworth:
> It looks like if you use `inherit` on the main posts query

The problem with this is that when you use `inherit` the query block seems to ignore other query parameters like `perPage`. I'm not sure if that's expected behaviour or a bug.

ryelle:
That's a good question. It looks like [the block query gets merged into the main query](https://github.com/WordPress/gutenberg/blob/9d4474c1c4c486947d479ecb817c4303ba8dfa86/packages/block-library/src/post-template/index.php#L29), but values in the main query will override the block values… I think that is intended. If you try this through the editor, hitting "Inherit query from template" turns off all the UI controls for taxonomies and per-page settings.

tellyworth:
Thanks for digging that up @ryelle.  I'm pretty sure the code in `render_block_core_post_template()` has a number of undesirable side effects that are going to cause problems for block themes. As well as the `perPage` thing noted above, this means you get inconsistent versions of query vars in filters. I think I'll have to write up some detailed test cases to clarify and figure out how to move forward.

PR #81, opened 2021-11-18
Rough starting template for month in WP

### LINKED ISSUE #77: month in wp - rough markup
See https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=4659%3A8683, "Category: Month in WP".

tellyworth:
This is a starting point for the month-in-wordpress templates. There's no CSS changes yet, just the markup.

Some notes:
* The wp:query block has some limitations and oddities. I've set columns to 4 because otherwise it was only showing 2. Could be an off-by-one bug or just some quirky layout stuff.
* I used perPage:10000. 0/-1 don't work like WP_Query. I'd prefer to set the limit to something like 60 (5 years per page), but it seems that the pagination block doesn't pay attention to the query block's page parameters and that would mean putting a magic number in two places. Probably there's a sensible way to handle that, but for now I've just removed the pagination altogether.

Fixes #77.

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/block-template-parts/content-category-month-in-wordpress.html
+++ b/source/wp-content/themes/wporg-news-2021/block-template-parts/content-category-month-in-wordpress.html
@@ -1,0 +1,11 @@
+<!-- wp:group {"tagName":"header","className":"entry-header"} -->
+<header class="wp-block-group entry-header">
+	<h2><!-- wp:post-date {"format":"M"} /--></h2>
+
+	<!-- wp:group {"className":"entry-meta"} -->
+	<div class="wp-block-group entry-meta">
+		<!-- wp:post-date {"format":"Y"} /-->
+	</div>
+	<!-- /wp:group -->
+</header>
+<!-- /wp:group -->
--- a/source/wp-content/themes/wporg-news-2021/block-template-parts/content-category-month-in-wordpress.html
+++ b/source/wp-content/themes/wporg-news-2021/block-template-parts/content-category-month-in-wordpress.html
@@ -1,6 +1,6 @@
 <!-- wp:group {"tagName":"header","className":"entry-header"} -->
 <header class="wp-block-group entry-header">
-	<h2><!-- wp:post-date {"format":"M"} /--></h2>
+	<h2><!-- wp:post-date {"format":"M","isLink":true} /--></h2>
 
 	<!-- wp:group {"className":"entry-meta"} -->
 	<div class="wp-block-group entry-meta">
--- a/source/wp-content/themes/wporg-news-2021/block-templates/category-month-in-wordpress.html
+++ b/source/wp-content/themes/wporg-news-2021/block-templates/category-month-in-wordpress.html
@@ -1,0 +1,16 @@
+<!-- wp:template-part {"slug":"header","align":"full","className":"site-header-container"} /-->
+<!-- This is a hack to ensure content isn't covered by the fixed header -->
+<!-- wp:template-part {"slug":"header","align":"full","className":"site-header-offset"} /-->
+
+<!-- wp:query {"tagName":"main","className":"site-content-container","displayLayout":{"type":"flex","columns":4},"query":{"perPage":"10000","inherit":false}} -->
+<main class="wp-block-query site-content-container">
+	<!-- wp:post-template -->
+		<!-- wp:template-part {"slug":"content-category-month-in-wordpress","tagName":"article","layout":{"inherit":true}} /-->
+	<!-- /wp:post-template -->
+
+</main>
+<!-- /wp:query -->
+
+<!-- wp:template-part {"tagName":"footer","slug":"footer-archive","className":"footer-archive","layout":{"inherit":true}} /-->
+
+<!-- wp:wporg/global-footer /-->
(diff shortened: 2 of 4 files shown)


source/wp-content/themes/wporg-news-2021/block-templates/category-month-in-wordpress.html
```diff
@@ -0,0 +1,16 @@
+<!-- wp:template-part {"slug":"header","align":"full","className":"site-header-container"} /-->
+<!-- This is a hack to ensure content isn't covered by the fixed header -->
+<!-- wp:template-part {"slug":"header","align":"full","className":"site-header-offset"} /-->
+
+<!-- wp:query {"tagName":"main","className":"site-content-container","displayLayout":{"type":"flex","columns":4},"query":{"perPage":"10000","inherit":false}} -->
```
iandunn: I'm not sure of a good way to put the blue paintbrush strokes in-between the years 🤔 

We can't group the posts into years and then loop through each group, since Gutenberg doesn't support dynamic templates - https://github.com/WordPress/gutenberg/issues/32939

_Aside: That one issue is responsible for 70% of the difficulties I've run into :/ . I've been tracking them at https://github.com/WordPress/wporg-news-2021/issues?q=is%3Aissue+is%3Aopen+label%3A%22blocked%3A+external%22, so it'd be good to add this one as well if we don't figure out a solution._

We could maybe try something with CSS, like

```css
body.category.month-in-wordpress .entry-header nth-of-type(12)::after {
    background-image: url(...)
}
```

...but that would need to account for how the current won't have 12 months, and previous years won't always have 12 months.

iandunn: At `1300px` some of the columns jump off the screen. 

I'm guessing some of the CSS alignment that's applied to typical categories is causing that. some of the containers might need `alignwide` to avoid that.

you might also need to change some of the alignment/post-content styles to apply to `body.category:not(.month-in-wordpress)` or something. 

I'm happy to dig in more if you run into any questions about how it's currently setup or anything.

source/wp-content/themes/wporg-news-2021/block-template-parts/content-category-month-in-wordpress.html
```diff
@@ -0,0 +1,11 @@
+<!-- wp:group {"tagName":"header","className":"entry-header"} -->
+<header class="wp-block-group entry-header">
+	<h2><!-- wp:post-date {"format":"M"} /--></h2>
+
+	<!-- wp:group {"className":"entry-meta"} -->
+	<div class="wp-block-group entry-meta">
+		<!-- wp:post-date {"format":"Y"} /-->
```
iandunn, line 7: This'll need to be a link, but I don't think there's a block that would create one w/ the post date as anchor text, so you might have to use the HTML block and do it manually

source/wp-content/themes/wporg-news-2021/block-templates/category-month-in-wordpress.html
iandunn: I think the query will also need `categoryIDs` to limit it to just the MIW cat.

source/wp-content/themes/wporg-news-2021/block-templates/category-month-in-wordpress.html
```diff
@@ -0,0 +1,16 @@
...
+<main class="wp-block-query site-content-container">
+	<!-- wp:post-template -->
+		<!-- wp:template-part {"slug":"content-category-month-in-wordpress","tagName":"article","layout":{"inherit":true}} /-->
+	<!-- /wp:post-template -->
+
```
iandunn, line 10: I'm guessing we'd want to keep the query navigation template part. 

The links are broken at the moment, but that'll be fixed in #70 and #80

iandunn (commented):
Looks like good progress 👍🏻 

This is also a good example of some of the typical limitations we've run into trying to implement a real-world design.

source/wp-content/themes/wporg-news-2021/block-templates/category-month-in-wordpress.html
iandunn: _( er, I meant these to be 3 separate threads, but all related to the query block. I guess GH combines them if they're for the same line 🙁  )_

source/wp-content/themes/wporg-news-2021/block-templates/category-month-in-wordpress.html
iandunn: We could leave it off for this template, since there's only ~100 posts, but we'll have to get it working for other templates, and I'm guessing the same solution would work here too

source/wp-content/themes/wporg-news-2021/block-templates/category-month-in-wordpress.html
tellyworth: After tinkering a bit, the best idea I have for the spacer between years is with a PHP filter that hooks `post_class`, and adds conditional classes to the post for various special contexts that we need. For instance a `first-in-year` or `last-in-year` class. Then use `::after` to deal with it in CSS.

A similar hack might work for other situations where conditional formatting is needed but the logic is too complex for CSS. Like the conditional blue backgrounds for #76.

It's not a great solution but would move things along. Do you see any problems with that?

iandunn: Yeah, that that sounds like a good workaround 👍🏻

source/wp-content/themes/wporg-news-2021/block-template-parts/content-category-month-in-wordpress.html
tellyworth: After much searching, it turns out that `<!-- wp:post-date {"isLink":true} /-->` does the job.

tellyworth:
This isn't complete but it might be possible to polish it up in time for the demo.

PR #88, opened 2021-11-26
Add community rough templates

tellyworth:
This adds a very rough version of the Community category page.

There's lots that doesn't look right or work properly yet, but it's a starting point to work on.


source/wp-content/themes/wporg-news-2021/functions.php
```diff
@@ -231,3 +234,95 @@ function clarify_body_classes( $classes ) {
...
+ *
+ * @return array
+ */
+function specify_post_classes( $classes, $extra_classes, $post_id ) {
+	$classes[] = 'post-year-' . get_the_date('Y');
+
+	global $wp_query;
+
+	// Add first-in-year and last-in-year to help put design elements in between year groups in the Month In WordPress category
+	if ( is_object( $wp_query ) && $wp_query->post_count > 1 ) {
```
iandunn, line 251: I wonder if the `post-year-{date}` is enough to achieve this? Maybe something like this for the SVG "separator" image between years?

```css
[class*="post-year-"]:last-of-type {
    content: "";
    background-image: url( 'images/foo.svg' );
}
```

Since `post-year-` is targeted with a wildcard, it should match the last instance of `post-year-2021`, `post-year-2020`, etc.

tellyworth: Yeah that might work, thanks! I'll leave in the first-in-year code in place for now with a note to remove it if it's not ultimately needed.

PR #90, opened 2021-11-26
Refine background for local nav banner

### LINKED ISSUE #67: Refine background for local nav & query title banner
#63 simplified how the local nav bar background/strokes work, so that we don't have to manually create new images for every screen that has a unique color.

It works, but still needs some iteration:

- [x] When the category list is toggled on, the mask covers the names of some categories, making them hard to read. `mask-size: cover` is probably to blame, but I couldn't find a way to change that and still get the desired effect when the category list is closed.
- [ ] at some breakpoints, the left/right edges don't look close to the mockup

The solution might be to replace `mask-image` with a `background-image` that's the inverse of the current one, like originally discussed in https://github.com/WordPress/wporg-news-2021/issues/62#issue-1039020046

It might also be worth considering multiple background images again, but in a way that doesn't require the duplicate images. Like one for the bottom tear, and one for the left/right edge tears. Or maybe combining a background image with a mask image. If 2 images are used, it might help to apply one to `nav.local-header` and the other to `div.local-header__navigation`

See #41 for details about the original implementation.

Related: #69, #84

MaggieCabrera:
This PR refactors the local-header banner to fix the issues outlined in https://github.com/WordPress/wporg-news-2021/issues/67

Now instead of resizing the SVG to fit the content of the dropdown, we are appending it to the bottom of the container inside the `::after` pseudo-element. To keep the background with the correct color for the rest of the dropdown while it's open I'm using a `linear-gradient` that starts right after the SVG.

I think I've covered all cases from the related issue, let me know if I've missed anything.

Screen Capture on 2021-11-26 at 15-42-55

Screenshot 2021-11-26 at 15 35 49

Screenshot 2021-11-26 at 15 35 30

Closes https://github.com/WordPress/wporg-news-2021/issues/67

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/components/_local-header.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/components/_local-header.scss
@@ -225,7 +225,7 @@
 %bottom-mask {
 	position: relative;
 	background-color: transparent;
-    background: linear-gradient(to top, transparent 58px, var( --bar-background-color ) 58px 100%);
+    background: linear-gradient(to top, transparent 56px, var( --bar-background-color ) 56px 100%);
 
 	&:after {
 		content: "";
--- a/source/wp-content/themes/wporg-news-2021/sass/components/_local-header.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/components/_local-header.scss
@@ -237,7 +237,6 @@
 		z-index: -1;
 		mask-image: url('images/local-nav-mask.svg');
 		mask-repeat: no-repeat;
-		mask-position-x: -20px;
 		mask-size: cover;
 		background-color: var( --bar-background-color );
 	}
--- a/source/wp-content/themes/wporg-news-2021/sass/components/_local-header.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/components/_local-header.scss
@@ -1,14 +1,3 @@
-%bottom-mask {
-	mask-image: url('images/local-nav-mask.svg');
-	mask-repeat: no-repeat;
-	mask-position: -200px -1px;
-	mask-size: cover;
-
-	@include break-small {
-		mask-position-x: -20px;
-	}
-}
-
 // Default/light version.
 .local-header {
 	--bar-text-color: var(--wp--preset--color--white);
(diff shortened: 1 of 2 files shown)


source/wp-content/themes/wporg-news-2021/sass/components/_query-title-banner.scss
```diff
@@ -26,7 +26,10 @@ body.search,
 body.archive:not(.category) {
```
iandunn, line 27: I noticed a few pages where `header.query-title-banner` is accidentally hidden, because the linear gradient removed the background color. The above selector might be a way to target all of them at once (but I haven't tested that).

- http://wporg-news.test/page/2/
- http://wporg-news.test/?s=george
- http://wporg-news.test/2020/
- http://wporg-news.test/author/iandunn/
- http://wporg-news.test/tag/5-8/

source/wp-content/themes/wporg-news-2021/sass/components/_local-header.scss
```diff
@@ -231,3 +220,25 @@
...
+		height: 58px;
+		position: absolute;
+		bottom: 0;
+		left: 0;
+		right: 0;
+		z-index: -1;
+		mask-image: url('images/local-nav-mask.svg');
+		mask-repeat: no-repeat;
+		mask-position-x: -20px;
+		mask-size: cover;
```
iandunn, line 240: It looks like the image doesn't stretch infinitely, it stops at around `2000px`, so there's just negative space after that.

Screen Shot 2021-11-26 at 9 23 03 AM

IIRC that was true before this PR, so feel free to open a new issue for it if you feel like it's not a quick iteration.

iandunn (approved):
Great idea about using `::after`! 

I think this is good to merge once the `query-title-banner` issue is fixed.

source/wp-content/themes/wporg-news-2021/sass/components/_query-title-banner.scss
tellyworth: @MaggieCabrera is this an easy fix?

source/wp-content/themes/wporg-news-2021/sass/components/_local-header.scss
MaggieCabrera: yeah I think it was the max position, should be fine now

source/wp-content/themes/wporg-news-2021/sass/components/_query-title-banner.scss
MaggieCabrera: yes, I think so, on it!

MaggieCabrera: ok so what I did is a bit hacky, I added a `:before` that hides the mask only when the menu is closed for those cases, that way the mask is still visible when the menu is open. It's more consistent with the rest of the site that way. Are these always blue? because I'm making that assumption here.

iandunn: I think that's a safe assumption, at least for now 👍🏻

PR #91, opened 2021-11-29
Refactor query pagination background strokes

### LINKED ISSUE #84: Refine background for `query-pagination` block
`query-pagination` has the same problem/needs as #67 . Once a solution is found there, it should be applied to this block too.

MaggieCabrera:
This PR implements the strokes for the query pagination using mask-image and after so that we can control the color of the strokes. I didn't use a different SVG for the blue background "all releases" block because I see you all want to avoid loading too many images.

I went ahead and added the stroke to the "News" heading too, let me know if that's not needed and I can remove that commit.

Screen Capture on 2021-11-29 at 10-43-57

Screenshot 2021-11-29 at 10 44 48

Screenshot 2021-11-29 at 10 44 42

Screenshot 2021-11-29 at 11 20 12

Closes https://github.com/WordPress/wporg-news-2021/issues/84


pbking:
Currently works great in Firefox.  Chrome is having som eissues with `mask-*` properties.

MaggieCabrera:
Looks like Chrome needs the -webkit prefix for it to work, I'll see if I can change that on the build tool

MaggieCabrera:
I added autoprefixer to the build script, after installing dependencies and rebuilding the CSS Chrome shows fine now.

Issue #100, opened 2021-12-02
Visual polish of bottom banner for SotW


tellyworth:
The bottom banner needs to be tested for SotW readiness, and any gliches raised as issues here. This is the block containing "follow the code", "find an event" etc. A recent screenshot from the test site:
<img width="1395" alt="Screen Shot 2021-12-02 at 3 41 22 pm" src="https://user-images.githubusercontent.com/7200686/144359016-d191c376-d7ff-4622-928e-f08fc2a109d9.png">

Figma for reference: https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=4659%3A8683

danieldudzic:
The h4 headings have the `font-size` set to `30px` on the test page - Figma has them at `26px`.

tellyworth:
Thanks @danieldudzic, well spotted. Do you have any opinions on whether #105 is the right way to fix that?

danieldudzic:
> Thanks @danieldudzic, well spotted. Do you have any opinions on whether #105 is the right way to fix that?

Good idea - `h5` already has the correct `font-size` - let's use it.

tellyworth:
Fixed in #105.

tellyworth:
Any other issues here or can we close this?

mikachan:
I've just tested this out and the only thing I can spot is the line height is slightly different between Figma and the test site. In Figma it's set to `30.4px` and on the site it's `25.6px` (from `var(--wp--custom--body--typography--line-height)`).

Otherwise, this is looking good to me, and it works well at all resolutions.

mikachan:
I've just noticed something else! The 'Listen to all episodes' link has left padding at mobile resolution, and on Figma it doesn't. And in the same area, it looks like the headings should be a larger font size at mobile: they're `26px` on Figma, compared to `20px` on the test site. The spacing below the headings also looks like it should be increased on the test site:

| Figma      | Test site |
| ----------- | ----------- |
|  ![image](https://user-images.githubusercontent.com/1645628/144897540-aa4e1dd5-e2ed-4fa3-a5fb-fe2ae80cded9.png)    | ![image](https://user-images.githubusercontent.com/1645628/144897413-3b6d6d8f-86ee-439e-b3f8-b1929abfea53.png)       |

mikachan:
I've opened a PR for the above issues: https://github.com/WordPress/wporg-news-2021/pull/107

scruffian:
[...]

PR #105, opened 2021-12-03
Use h5 instead of h4

tellyworth:
The headings are supposed to be 26px, which is what h5 is.

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/block-template-parts/footer-archive.html
+++ b/source/wp-content/themes/wporg-news-2021/block-template-parts/footer-archive.html
@@ -3,7 +3,7 @@
 	<!-- wp:group {"className":"follow-the-code"} -->
 	<div class="wp-block-group follow-the-code">
 		<!-- wp:heading {"level":4} -->
-		<h4>Follow The Code</h4>
+		<h5>Follow The Code</h5>
 		<!-- /wp:heading -->
 
 		<!-- wp:paragraph -->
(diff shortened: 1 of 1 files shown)


tellyworth:
See https://github.com/WordPress/wporg-news-2021/issues/100#issuecomment-985157996

danieldudzic (approved):
Works fine. I'm comfortable with this fix 👍

source/wp-content/themes/wporg-news-2021/block-template-parts/footer-archive.html
```diff
@@ -3,7 +3,7 @@
 	<!-- wp:group {"className":"follow-the-code"} -->
 	<div class="wp-block-group follow-the-code">
 		<!-- wp:heading {"level":4} -->
-		<h4>Follow The Code</h4>
+		<h5>Follow The Code</h5>
```
ryelle, line 6: The attribute here should also be updated, otherwise you'll get a broken block in the editor - the `level` is used to determine which `h*` to render.

ryelle (commented):
Using different heading levels just for visual tweaks breaks the semantics of the page, which is important for screen reader users - [headings shouldn't jump levels](https://webaim.org/techniques/semanticstructure/#headings).

Right now, the page hierarchy looks like this, with the h5s following an h2 People of WP. There should probably be another (visually hidden) h2 for the footer section, and these should be h3.

Screen Shot 2021-12-06 at 5 18 47 PM

ryelle:
(I know this PR is merged, it seemed like a more relevant place to comment than the issue)

tellyworth:
Thanks! Let's open a fresh issue to fix this after SotW.

danieldudzic:
> Using different heading levels just for visual tweaks breaks the semantics of the page, which is important for screen reader users - [headings shouldn't jump levels](https://webaim.org/techniques/semanticstructure/#headings).
> 
> Right now, the page hierarchy looks like this, with the h5s following an h2 People of WP. There should probably be another (visually hidden) h2 for the footer section, and these should be h3.
> 
> Screen Shot 2021-12-06 at 5 18 47 PM

Thank you for highlighting the accessibility aspect.

Issue #103, opened 2021-12-02
Categories dropdown has no focus state


ryelle:
When keyboard-navigating through the header, after the nav menu, the focus disappears. If I hit space or enter the categories list toggles, so I assume I'm on the hidden checkbox here.

<img width="270" alt="Screen Shot 2021-12-02 at 4 04 44 PM" src="https://user-images.githubusercontent.com/541093/144502982-591263c6-8ac2-4366-95d0-cad655b51edd.png">

This should get a focus state so it's clear, it can be done with a similar sibling selector:

```css
.local-header__categories-toggle:focus  ~ label {
    outline: 1px dotted currentColor;
}
```

tellyworth:
I made a PR for this. It needs some styling (see screenshot) and probably further work, but finishing it can wait till after SotW.
<img width="210" alt="Screen Shot 2021-12-06 at 5 17 52 pm" src="https://user-images.githubusercontent.com/7200686/144797104-100173b4-4b6c-4e5e-b16b-fe3ee697ace0.png">

ryelle:
Fixed in #106

PR #106, opened 2021-12-06
Try outline focus state

tellyworth:
WIP fix for #103. Needs styling.


ryelle:
I moved the focus style to the span, so it's not awkwardly tall 🙂 

Screen Shot 2021-12-07 at 4 45 03 PM

Maybe post SOTW that area can be revisited, since the focus should match the target area, but it's not a problem.

ryelle (approved):

PR #107, opened 2021-12-07
SotW bottom banner improvements

mikachan:
This PR adds a few tweaks to enhance the bottom banner:

- Refactor the podcast link spacing so the 'Listen to all episodes' link sits to the left at small resolutions
- Increases the line height of the bottom banner in line with design in Figma
- Adds spacing below headings in line with Figma
- Increases font size of headings at mobile in line with Figma

| Before | After |
| ------- | -------- |
| image | image |

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/block-template-parts/footer-archive.html
+++ b/source/wp-content/themes/wporg-news-2021/block-template-parts/footer-archive.html
@@ -2,8 +2,8 @@
 <div class="wp-block-group alignfull footer-archive-row">
 	<!-- wp:group {"className":"follow-the-code"} -->
 	<div class="wp-block-group follow-the-code">
-		<!-- wp:heading {"level":4} -->
-		<h5>Follow The Code</h5>
+		<!-- wp:heading {"level":3} -->
+		<h3>Follow The Code</h3>
 		<!-- /wp:heading -->
 
 		<!-- wp:paragraph -->
@@ -14,8 +14,8 @@
 
 	<!-- wp:group {"className":"find-an-event"} -->
 	<div class="wp-block-group find-an-event">
-		<!-- wp:heading {"level":4} -->
-		<h5>Find An Event Near You</h5>
+		<!-- wp:heading {"level":3} -->
+		<h3>Find An Event Near You</h3>
 		<!-- /wp:heading -->
 
 		<!-- wp:paragraph -->
@@ -30,8 +30,8 @@
 <div class="wp-block-group alignfull footer-archive-row">
 	<!-- wp:group {"className":"subscribe-wp-news"} -->
 	<div class="wp-block-group subscribe-wp-news">
-		<!-- wp:heading {"level":4} -->
-		<h5>Subscribe to WordPress News</h5>
+		<!-- wp:heading {"level":3} -->
+		<h3>Subscribe to WordPress News</h3>
 		<!-- /wp:heading -->
 
 		[jetpack_subscription_form show_only_email_and_button="true" show_subscribers_total="true"]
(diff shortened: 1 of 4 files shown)


beafialho:
The "After" screenshot looks good to me 👍

ryelle:
Can you change the headings back to `h3`? they were changed to `h5`s in this PR #105, but it breaks the heading semantics. I mentioned in [that review that](https://github.com/WordPress/wporg-news-2021/pull/105#pullrequestreview-824574598)

> There should probably be another (visually hidden) h2 for the footer section, and these should be h3.

mikachan:
> Can you change the headings back to h3?

I've changed the headings to h3s, and I've added a hidden h2 with the text, "Get the Latest Updates".

I've made the headings `26px` at mobile, but I left the desktop size at the default for h3s, which is `36px`. Is this OK, or should I reduce it back to `26px`? Here's what the larger size looks like:

image

source/wp-content/themes/wporg-news-2021/block-template-parts/footer-archive.html
```diff
@@ -1,72 +1,80 @@
...
-	<!-- wp:group {"className":"follow-the-code"} -->
-	<div class="wp-block-group follow-the-code">
-		<!-- wp:heading {"level":4} -->
-		<h5>Follow The Code</h5>
-		<!-- /wp:heading -->
+<!-- wp:group {"align":"full","classname":"footer-archive"} -->
+<div class="wp-block-group alignfull footer-archive">
+	<!-- wp:heading {"level":2} -->
+	<h2>Get the Latest Updates</h2>
+	<!-- /wp:heading -->
```
ryelle:
```suggestion
	<!-- wp:heading {"level":2,"classname":"screen-reader-text"} -->
	<h2 class="screen-reader-text">Get the Latest Updates</h2>
	<!-- /wp:heading -->
```

You can use the default screen-reader-text class here to hide this

source/wp-content/themes/wporg-news-2021/sass/components/footer-archive/_footer-archive.scss
```diff
@@ -1,32 +1,47 @@
...
-			border-top: 1px solid var(--wp--preset--color--light-grey);
-			border-bottom: none;
-
-			&:nth-child(odd) {
-				border-right: 1px solid var(--wp--preset--color--light-grey);
-				padding-right: var( --wp--custom--alignment--edge-spacing );
-			}
+.footer-archive {
+	h2 {
+		display: none;
```
ryelle: `display: none` actually hides this content from screen readers too. You need to do [a bit more to visually hide content](https://github.com/WordPress/gutenberg/blob/3475693ae69d3464defa842a76da8505bc275a46/packages/block-library/src/common.scss#L62), which is why the `screen-reader-text` class helper exists 🙂

ryelle (changes requested):
> I've made the headings 26px at mobile, but I left the desktop size at the default for h3s, which is 36px.

I think it's supposed to be 26px in both places, according to the design.

mikachan:
Brill thanks, I didn't realise the `screen-reader-text` class existed.

I've added a specific size for these headings just for the footer, so they're always `26px`. Ready for another review, thanks for your help!

danieldudzic:
Changes look fine. I just have one question: There seems to be a discrepancy (in Figma) in the spacing between the heading and paragraph in the bottom banner:
px

Not sure which one is correct and should be used.

Requesting guidance from @beafialho :)

beafialho:
Sure, we can make them all `10px`

source/wp-content/themes/wporg-news-2021/sass/components/footer-archive/_footer-archive.scss
ryelle: This `display:none` still needs to be removed, it's hiding the heading:

Screen Shot 2021-12-08 at 10 36 53 AM

ryelle (changes requested):
Looking good visually, just [one more request](https://github.com/WordPress/wporg-news-2021/pull/107#discussion_r764979856) 🙂

mikachan:
Gah sorry, removed this now! Thanks @ryelle.

ryelle (approved):
Looks good now 👍🏻

ryelle:
Merging this, if we want to adjust the margins, it can happen in another PR.

mikachan:
The margins mentioned above are already `10px`, so I think everything discussed here has been actioned.

Issue #83, opened 2021-11-19
State of the Word preview


iandunn:
This is an overview of [the milestone](https://github.com/WordPress/wporg-news-2021/milestone/6). The goal is to get critical parts of the site ready to demo at the State of the Word. New issues can be broken off of here when necessary.

### Desktop:

1. open browser, maximize it (should be `1320px` wide)
1. open w.org/news-test
1. open local nav categories dropdown
1. close local nav categories dropdown
1. scroll down to "latests posts" section
1. click on a post (pick one a specific one few days before)
1. scroll through that post

### Mobile:

1. click "News" in local nav bar to go back to homepage
1. open DevTools
1. click responsive mode button
1. Choose `?` from the menu (should be `375px` wide)
1. Scroll through homepage and single post like you did for mobile

There won't be any interactions with global search, hover states, clicking any other links, etc. it's just the the visual aspects of the above that we're concerned with

### Visual Polish
The following templates need to be visually polished to match [the mockup](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=4659%3A8682). They don't need to be any more functionally complete than they already are, though.

- [ ] global header - #86
- [ ] local header - #86, #90
- [x] #96
    - [x] people of wp images flex layout needs to fill width at all viewports - maybe use something similar to "priority navigation" from #86
    - [x] Enable subscription #35
    - [x] Add podcast icons directly to theme for now, see #36 
- [x] #97
     - [x] Hide likes/sharing if #24 isn't done in time
- [x] #98
- [x] #100
- [x] #101
- [x] #102

### Additional tasks:

- [x] import content #82 
- [x] import new posts the friday before the event, so that posts match production
- [ ] sync production w/ latest code as needed

### Demo video

The demo is to be pre-recorded (by who?)

- [ ] Refine the exact sequence of clicks and views based on above (including decisions about which posts and categories to show)
- [ ] Write a script
- [ ] Record a test video
- [ ] Check the video closely and raise issues for anything that needs fixing or refining
- [ ] Record the final video

PR #126, opened 2021-12-13
Quick temporary fix for See All Releases and See All People

tellyworth:
I'm not sure how to get the query-pagination-next blocks to work as expected.

This isn't a permanent fix, and doesn't yet cover the All Posts view (where should that live?)

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/block-template-parts/front-page/latest-posts.html
+++ b/source/wp-content/themes/wporg-news-2021/block-template-parts/front-page/latest-posts.html
@@ -21,7 +21,7 @@
 	<!-- /wp:post-template -->
 
 	<!-- wp:query-pagination -->
-		<!-- wp:query-pagination-next {"label":"See All Posts"} /-->
+	<a href="./category/uncategorized/" class="wp-block-query-pagination-next">See All Posts</a>
 	<!-- /wp:query-pagination -->
 </section>
 <!-- /wp:query -->
--- a/source/wp-content/themes/wporg-news-2021/block-template-parts/front-page/latest-posts.html
+++ b/source/wp-content/themes/wporg-news-2021/block-template-parts/front-page/latest-posts.html
@@ -21,7 +21,7 @@
 	<!-- /wp:post-template -->
 
 	<!-- wp:query-pagination -->
-	<a href="./category/uncategorized/" class="wp-block-query-pagination-next">See All Posts</a>
+	<a href="./page/2/" class="wp-block-query-pagination-next">See All Posts</a>
 	<!-- /wp:query-pagination -->
 </section>
 <!-- /wp:query -->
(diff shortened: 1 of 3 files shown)


source/wp-content/themes/wporg-news-2021/block-template-parts/front-page/latest-posts.html
```diff
@@ -21,7 +21,7 @@ <h2 class="front__latest-posts-heading">Latest Posts</h2>
 	<!-- /wp:post-template -->
 
 	<!-- wp:query-pagination -->
-		<!-- wp:query-pagination-next {"label":"See All Posts"} /-->
+	<a href="./category/uncategorized/" class="wp-block-query-pagination-next">See All Posts</a>
```
tellyworth: Open to suggestions for a quick hacky workaround we could use here for SotW - Uncategorized has the right effect but doesn't include all posts.

iandunn:
I don't think clicking on those links is part of [the demo script](https://github.com/WordPress/wporg-news-2021/issues/83), is it?

If not, we could avoid a short-term fix like this, and then have time to dig into a long-term fix.

tellyworth:
What do you think about directly using `a` tags for Releases and People as a long-term fix? It's simple and semantic. The only hacky part with the current PR is that I put it inside the `wp:query-pagination` block in order to avoid touching the CSS. We'd need a way to translate it though.

The All Posts link is definitely a hack and we need a proper solution for that either way.

iandunn:
It looks like @ryelle [figured out the solution](https://github.com/WordPress/wporg-news-2021/issues/70#issuecomment-993764835). That feels better to me than hardcoding URLs, since it'll change if needed, and is more semantic.

It's not a huge deal if y'all feel like this is better, though.

ryelle:
That'll only work on the main "All Posts" link, since it inherits the default WP query. These other two (releases & people) will still need to be hardcoded, since they're additional queries.

iandunn:
👍🏻

tellyworth:
I think once #179 is merged this should pretty much solve the last of the pagination issues. Might need a light refresh.

shaunandrews:
I'm not entirely sure, but I _think_ this PR aims to resolve the broken "See All Posts" link on the front-page `/`. If so, that in my testing it does work. However, with the title of "quick temporary fix" I'm left wondering if this should be closed and fixed in a better way...

Is this PR still relevant?

ryelle:
It's intending to fix "See All Posts", "See All Releases", and "See All People". It's a quick fix because it's kind of hacky, using the pagination block comment but putting a link inside will definitely break in the editor.

I think a better approach would be just using a group or paragraph block with a class & a link, that might need some style tweaking though.

ryelle:
Closing in favor of #203.

Issue #120, opened 2021-12-10
Global & Local Header Improvements


beafialho:
Here are some visual improvements.

**Desktop:**

- [x] There should be a dark overlay on the page when the categories menu is open
<img width="1413" alt="Captura de ecrã 2021-12-10, às 10 09 36" src="https://user-images.githubusercontent.com/65220155/145556843-5f093310-c6f3-48ea-989c-1baf1705c558.png">

- [x] Global header looking off in larger breakpoints ([reference on how it should look](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=7413%3A31250))
<img width="1212" alt="Captura de ecrã 2021-12-10, às 10 09 49" src="https://user-images.githubusercontent.com/65220155/145556829-16f71839-8c03-4c8b-a183-0c5f3a1b403b.png">

- [x] Currently, there’s quite some space left in the global header menu. The rule should be: if there is enough space for all the menu links, it shouldn’t collapse into the `...` Seems like the margin is too big on the right, it can be just `25px` like in its left margin
<img width="1312" alt="Captura de ecrã 2021-12-10, às 10 12 54" src="https://user-images.githubusercontent.com/65220155/145557223-18f08865-8cf7-4980-b2b4-11f3999511b1.png">

**Mobile:**
- [x] Categories menu should be right aligned

| Test Site      | Mockup |
| ----------- | ----------- |
| <img width="343" alt="Captura de ecrã 2021-12-10, às 09 42 45" src="https://user-images.githubusercontent.com/65220155/145557515-a49ef3ab-27ba-4934-b55a-db623a876c89.png"> | <img width="252" alt="Captura de ecrã 2021-12-10, às 10 15 40" src="https://user-images.githubusercontent.com/65220155/145557633-5b7bcbb4-ff0d-42f8-a336-586b00380537.png">  |

- [x] When the Categories menu is open, the active item should be signaled and in bold.

| Test Site      | Mockup |
| ----------- | ----------- |
[...]

PR #127, opened 2021-12-13
Adding overlay when the local header is open

matiasbenedetto:
## Description
Adding overlay when the local header is open
fix part of https://github.com/WordPress/wporg-news-2021/issues/120

## Screenshots
**Before:**
image
image

**After:**
image
image

## Issue
https://github.com/WordPress/wporg-news-2021/issues/120

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/components/_local-header.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/components/_local-header.scss
@@ -132,6 +132,20 @@
 					svg {
 						transform: rotate(180deg);
 					}
+					
+					&::before {
+						content: "";
+						display: block;
+						position: fixed;
+						top: 0;
+						left: 0;
+						width: 100%;
+						height: 100vh;
+						background-color: rgba(0,0,0,0.5);
+						z-index: -1;
+						animation-duration: 400ms;
+						animation-name: overlay-fade-in;
+					}
 				}
 
 				~ .wp-block-categories {


iandunn:
In the After image, I see some grey background inside the transition area (between the light blue categories and the dark blue "News" site title).

I don't see an example in the mockup, but I assume that should be dark blue instead of grey, like it is for the desktop. Is it intended to be grey?

tellyworth:
I think that grey background in the mobile view is expected. There's some white background there in the before, and the grey is just white plus the transparent overlay.

I'd lean towards merging this and handling any improvements in a new issue.

iandunn:
👍🏻

beafialho:
> In the After image, I see some grey background inside the transition area (between the light blue categories and the dark blue "News" site title).

Thanks for addressing this @matiasbenedetto.

It all seems accurate to me, except for the mobile "After" screenshot, where it shouldn't have that white/grey background that @iandunn mentioned.

| Screenshot      | Intended |
| ----------- | ----------- |
|  Captura de ecrã 2021-12-15, às 15 23 51 | Captura de ecrã 2021-12-15, às 15 21 11 |

I also see it when the local nav bar is not open, in which case it should be similar to the desktop view and blend with the background.

| Test Site    | Intended |
| ----------- | ----------- |
| Captura de ecrã 2021-12-15, às 15 25 38 | Captura de ecrã 2021-12-15, às 15 27 03 |

PR #142, opened 2021-12-22
Use a filter to set posts_per_page for some categories

tellyworth:
This lets us use `inherit:true` for the `wp:query` blocks in the page templates. Without which we run into problems with pagination etc.

See #70 for discussion.

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/block-templates/category-community.html
+++ b/source/wp-content/themes/wporg-news-2021/block-templates/category-community.html
@@ -2,6 +2,7 @@
 <!-- This is a hack to ensure content isn't covered by the fixed header -->
 <!-- wp:template-part {"slug":"header","align":"full","className":"site-header-offset"} /-->
 
+<!-- Note: This query has filters, but they live in `category_posts_per_page()`. That's necessary to avoid a Gutenberg design flaw. -->
 <!-- wp:query {"tagName":"main","className":"site-content-container","query":{"inherit":true},"displayLayout":{"type":"flex","columns":4},"align":"full"} -->
 <main class="wp-block-query site-content-container">
 	<!-- wp:post-template -->
--- a/source/wp-content/themes/wporg-news-2021/block-templates/category-community.html
+++ b/source/wp-content/themes/wporg-news-2021/block-templates/category-community.html
@@ -2,7 +2,7 @@
 <!-- This is a hack to ensure content isn't covered by the fixed header -->
 <!-- wp:template-part {"slug":"header","align":"full","className":"site-header-offset"} /-->
 
-<!-- wp:query {"tagName":"main","className":"site-content-container","query":{"category":"community","perPage":"20","postType":"post","inherit":false},"displayLayout":{"type":"flex","columns":4},"align":"full"} -->
+<!-- wp:query {"tagName":"main","className":"site-content-container","query":{"inherit":true},"displayLayout":{"type":"flex","columns":4},"align":"full"} -->
 <main class="wp-block-query site-content-container">
 	<!-- wp:post-template -->
 		<!-- wp:template-part {"slug":"content-category-community","tagName":"article","layout":{"inherit":true}} /-->
--- a/source/wp-content/themes/wporg-news-2021/block-templates/category-month-in-wordpress.html
+++ b/source/wp-content/themes/wporg-news-2021/block-templates/category-month-in-wordpress.html
@@ -2,7 +2,7 @@
 <!-- This is a hack to ensure content isn't covered by the fixed header -->
 <!-- wp:template-part {"slug":"header","align":"full","className":"site-header-offset"} /-->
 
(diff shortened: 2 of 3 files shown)


source/wp-content/themes/wporg-news-2021/block-templates/category-community.html
```diff
@@ -2,7 +2,7 @@
 <!-- This is a hack to ensure content isn't covered by the fixed header -->
 <!-- wp:template-part {"slug":"header","align":"full","className":"site-header-offset"} /-->
 
-<!-- wp:query {"tagName":"main","className":"site-content-container","query":{"category":"community","perPage":"20","postType":"post","inherit":false},"displayLayout":{"type":"flex","columns":4},"align":"full"} -->
+<!-- wp:query {"tagName":"main","className":"site-content-container","query":{"inherit":true},"displayLayout":{"type":"flex","columns":4},"align":"full"} -->
```
iandunn, line 5:
```suggestion
<!-- Note: This query has filters, but they live in `category_posts_per_page()`. That's necessary to avoid a Gutenberg design flaw. -->
<!-- wp:query {"tagName":"main","className":"site-content-container","query":{"inherit":true},"displayLayout":{"type":"flex","columns":4},"align":"full"} -->
```

source/wp-content/themes/wporg-news-2021/block-templates/category-month-in-wordpress.html
```diff
@@ -2,7 +2,7 @@
 <!-- This is a hack to ensure content isn't covered by the fixed header -->
 <!-- wp:template-part {"slug":"header","align":"full","className":"site-header-offset"} /-->
 
-<!-- wp:query {"tagName":"main","className":"site-content-container","displayLayout":{"type":"flex","columns":4},"query":{"category":"month-in-wordpress","perPage":"10000","inherit":false}} -->
+<!-- wp:query {"tagName":"main","className":"site-content-container","displayLayout":{"type":"flex","columns":4},"query":{"inherit":true}} -->
```
iandunn, line 6:
```suggestion
<!-- Note: This query has filters, but they live in `category_posts_per_page()`. That's necessary to avoid a Gutenberg design flaw. -->
<!-- wp:query {"tagName":"main","className":"site-content-container","displayLayout":{"type":"flex","columns":4},"query":{"inherit":true}} -->
```

source/wp-content/themes/wporg-news-2021/block-templates/category-community.html
iandunn, line 6:
```suggestion
<!-- Note: This query has filters, but they live in `category_posts_per_page()`. That's necessary to avoid a Gutenberg design flaw. -->
<!-- wp:query {"tagName":"main","className":"site-content-container","query":{"inherit":true},"displayLayout":{"type":"flex","columns":4},"align":"full"} -->
```

source/wp-content/themes/wporg-news-2021/functions.php
```diff
@@ -213,6 +214,25 @@ function offset_paginated_index_posts( $query ) {
...
 
+/**
+ * Adjust posts_per_page separately for certain categories.
+ *
+ * We have certain category templates designed for different numbers of posts per page. Unfortunately it's not possible to set this in the block temlates.
+ * (See https://github.com/WordPress/wporg-news-2021/issues/70#issuecomment-996460735)
+ *
+ * @param WP_Query $query
+ */
+function category_posts_per_page( $query ) {
```
iandunn: We could give this function a more generic name, in case we need to add more filters in the future. Not a big deal though.

iandunn (approved):
It's unfortunate that we need a hack, but this seems like a good one 👍🏻 

I made a few suggestions, but they're not blockers.

It'd probably be helpful to add a comment to https://github.com/WordPress/gutenberg/issues/32531, or open a new G issue, to offer feedback on the problems we encountered.

source/wp-content/themes/wporg-news-2021/functions.php
tellyworth: Yeah good suggestion, I'll update it. Perhaps something like `override_category_query_args()`.

I did wonder if this should be combined with (or call?) `offset_paginated_index_posts()`, since they're somewhat similar and could interact. Probably simplest to keep them separate unless there's a specific reason to do otherwise.

iandunn: 👍🏻

source/wp-content/themes/wporg-news-2021/block-templates/category-month-in-wordpress.html
tellyworth: Yeah I totally agree about posting feedback upstream. I need to gather my thoughts and probably discuss in Slack to straighten things out.

source/wp-content/themes/wporg-news-2021/block-templates/category-community.html
iandunn: "limitation" is probably a more generous word than "flaw" in my suggestion, especially since it's a beta feature. it'd probably also be good to link to https://github.com/WordPress/gutenberg/issues/32531, so that it's clear what the limitation refers to

Issue #139, opened 2021-12-20
Homepage: CTAs incorrect placement


beafialho:
In the homepage "Latest Release" and "Latest Posts" sections:
- the CTA links should be horizontally aligned with the posts list, positioned above the stroke shapes
- the stroke shapes should look similar to the ones in the mockup

| Test Site      | Intended |
| ----------- | ----------- |
|  <img width="1530" alt="Captura de ecrã 2021-12-20, às 15 21 14" src="https://user-images.githubusercontent.com/65220155/146791166-463bd6fe-dd24-4bbc-8977-2fc88fe45b19.png">  | <img width="1115" alt="Captura de ecrã 2021-12-20, às 15 26 54" src="https://user-images.githubusercontent.com/65220155/146791571-7b6c3f52-b829-4578-b200-c7fa59d2245c.png">   |
| <img width="1529" alt="Captura de ecrã 2021-12-20, às 15 21 32" src="https://user-images.githubusercontent.com/65220155/146791156-d5fa755f-d90a-4929-a62d-c27e09ed0a84.png"> |  <img width="1114" alt="Captura de ecrã 2021-12-20, às 15 27 00" src="https://user-images.githubusercontent.com/65220155/146791603-65d17d15-d2cf-48e4-b073-11e937dadaf7.png">   |

tellyworth:
Should be fixed by #144. See screenshots:
<img width="1398" alt="Screen Shot 2022-01-04 at 3 21 28 pm" src="https://user-images.githubusercontent.com/7200686/148009297-5d09ab22-ffe4-40b2-9236-aed5a60cbd06.png">
<img width="1417" alt="Screen Shot 2022-01-04 at 3 20 58 pm" src="https://user-images.githubusercontent.com/7200686/148009312-05566ebd-e688-4e09-857f-449ad2dd21de.png">

beafialho:
Looks accurate to me, thanks.

tellyworth:
Fixed by #144

PR #144, opened 2022-01-04
Fix/cta alignment

tellyworth:
See #139

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/blocks/_query-pagination.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/blocks/_query-pagination.scss
@@ -4,7 +4,7 @@
 
 .wp-block-query-pagination {
 	display: inline-block;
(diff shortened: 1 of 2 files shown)


source/wp-content/themes/wporg-news-2021/sass/blocks/_query-pagination.scss
```diff
@@ -4,7 +4,7 @@
 
 .wp-block-query-pagination {
 	display: inline-block;
-	line-height: 7em; // stylelint-disable-line declaration-property-unit-allowed-list
+	line-height: normal; // stylelint-disable-line declaration-property-unit-allowed-list
```
iandunn: We might not need the `stylelint` comment after this.

iandunn (approved):
#139 is in the post-launch milestone, so I don't think this is a high priority, but there's no reason to block merge if it's ready.

If it also looks correct on mobile, then this LGTM 👍🏻

PR #160, opened 2022-01-07
Fix heading font sizes

### LINKED ISSUE #159: Headings size is looking much smaller across the whole site. See style sheet for correct sizes
Captura de ecrã 2022-01-04, às 10 55 18

tellyworth:
See https://github.com/WordPress/wporg-mu-plugins/issues/44#issuecomment-1004410980

Seems this needs to change to keep up with Gutenberg, if I'm understanding correctly. Bit of a stab in the dark that seems to work.

Fixes #159.


ryelle (approved):
Looks good 👍🏻

PR #162, opened 2022-01-07
Tools: Switch from node-sass to sass for building CSS.

ryelle:
Node-sass is a deprecated project, new projects should be using the dart-sass implementation. This is a simple swap-out of the package, no tweaks made to the build process.

This also removes a workaround for `min` & `max`, which is not necessary anymore.

To test:

- Reinstall dependencies with `yarn`
- Run the build task: `yarn workspace wporg-news-2021-theme build`
- There should be no visual changes


tellyworth:
WFM

PR #174, opened 2022-01-10
Design Fixes: Left-align footer navigation menus

### LINKED ISSUE #153: Local footer text is centered on Month in WordPress category page

allancole:
This PR left-aligns the footer navigation menus. 

Note: This fix overrides a style from an mu-plugin which can be found here: 
`/wporg-news-2021/source/wp-content/mu-plugins/wporg-mu-plugins/mu-plugins/blocks/global-header-footer/postcss/footer/footer.pcss`

Since that plugin isn’t part of this project, this override is necessary to correct the alignment. 

Before: 
image

After: 
image

Fixes: #153

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/components/footer-archive/_footer-archive.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/components/footer-archive/_footer-archive.scss
@@ -40,6 +40,11 @@
 			}
 		}
 	}
+
+	// Left-align footer navigation menus
+	.wp-block-group.global-footer .wp-block-navigation__container {
+		align-items: self-start;
+	}
 }
 
 // Alternate dark version.


beafialho:
Thanks @allancole, this looks accurate!

ryelle:
For context, I think this is caused by this gutenberg bug: https://github.com/WordPress/gutenberg/issues/37695

Unless these footer items should be centered on other wporg sites, this should be fixed in [the mu-plugins project](https://github.com/WordPress/wporg-mu-plugins). They're two separate repos, but work happens in both.

allancole:
Makes sense, @ryelle. 
Should I submit a separate PR to the mu-plugins project? (Although, I doubt I have access)

ryelle:
I ported this over in https://github.com/WordPress/wporg-mu-plugins/issues/64, so the root issue should be fixed now 👍🏻

allancole:
Amazing. Thanks @ryelle !

Issue #176, opened 2022-01-11
at `470px`, On the homepage (and probably some others), when the `Categories` dropdown is closed, the local nav background doesn't blend into the blue News section that follows it


tellyworth:
at 470px, On the homepage (and probably some others), when the Categories dropdown is closed, the local nav background doesn't blend into the blue News section that follows it

See #134, #120

shaunandrews:
I think this is fixed in #178

Issue #175, opened 2022-01-11
Local nav category wrap point adujstment


tellyworth:
at `378px`, `Categories` in the local nav drops down to a 2nd line, but it should be on the same as `News`. there's a point at which it will have to wrap, but it can go much farther than it does
(see #134, #120)

shaunandrews:
I think this is fixed in #178

Issue #172, opened 2022-01-10
In smaller screens, the local nav bar menu should be displayed vertically instead of horizontally


tellyworth:
In smaller screens, the local nav bar menu should be displayed vertically instead of horizontally, similar to the header menu on mobile:
<img width="302" alt="Captura de ecrã 2022-01-04, às 11 35 25" src="https://user-images.githubusercontent.com/65220155/148053299-a547ba32-00f2-4169-bb23-0da0a7561140.png">

shaunandrews:
This is fixed in #178

Issue #171, opened 2022-01-10
Same weird white background behind the local nav bar appearing in all browsers:


tellyworth:
on smaller screens
<img width="567" alt="Captura de ecrã 2022-01-04, às 11 30 55" src="https://user-images.githubusercontent.com/65220155/148052841-ef0f8a3a-d1d6-43ff-aff3-fe128d50731b.png">
<img width="498" alt="Captura de ecrã 2022-01-04, às 11 30 42" src="https://user-images.githubusercontent.com/65220155/148052845-1abf5401-aab8-4cba-969d-bd8d74eb26eb.png">
<img width="446" alt="Captura de ecrã 2022-01-04, às 11 30 07" src="https://user-images.githubusercontent.com/65220155/148052847-22e6546c-bcc4-4279-9c10-1a2af46a8678.png">

shaunandrews:
This is fixed in #178

PR #178, opened 2022-01-11
Local Header Bar: Fix various issues and remove markup duplication

shaunandrews:
The PR refactors the "local-header" addressing a handful of issues: #163 #184 #176 #175 #172 #171

--

- Fixes the extra white-space below the Top Bar and the homepage header by adding a color to the body tag.
- Uses Flexbox to avoid having a duplicate header template in every template. (No obvious visual change, but does change some behavior.)
- Ensures the right hover color for links in the Top Bar.
- Adds a transition to the scrim and chevron arrow when toggling the Categories menu.


iandunn:
There's a simpler solution to removing the white gap: https://github.com/WordPress/wporg-news-2021/issues/163#issuecomment-1009362372

I'd be awesome to remove the need for the duplicate markup, though. If it's not easy to finish that part, it may be better to go with the simple solution for now, and then circle back after the launch, since we're on a tight deadline.

shaunandrews:
> I'd be awesome to remove the need for the duplicate markup, though. If it's not easy to finish that part

I think that part _is_ finished. This could use a review.

source/wp-content/themes/wporg-news-2021/sass/components/_local-header.scss
```diff
@@ -1,130 +1,125 @@
...
-			@include break-xlarge {
-				padding: 0;
-			}
-		}
+		// This overrides a left/right padding using
+		// the `--wp--custom--alignment--edge-spacing` variable.
+		//
+		// The !important is temporary and should  be accomplished
+		// with more specific selectors.
+		// padding: 0 !important;
```
ryelle: Was this not necessary after all?

source/wp-content/themes/wporg-news-2021/sass/components/_local-header.scss
```diff
@@ -133,104 +128,87 @@
...
-				@include break-mobile {
-					border: none;
-
-					&::before {
-						content: "\0020\00B7\0020";
-					}
+				position: relative;
+
+				// On small viewports the current category is highlighted
+				// using a right-side border.
```
ryelle, line 205: is this still in-progress?

ryelle (commented):
Overall this is looking great. 💯  for getting rid of the duplicate header.

There are a few places with commented-out code— if you're still working on that, fine, otherwise it should get removed.

I noticed a few issues in the small-screen menu:

- The padding to the right of the Categories label seems too big
- This should be a single column (#172, which is how it looks on trunk to me, so I don't know what that issue is pointing out)

Screen Shot 2022-01-11 at 3 00 38 PM

I also noticed that I can hover and interact with the page behind the overlay when the Categories menu is open, which is not the case on trunk.

source/wp-content/themes/wporg-news-2021/sass/components/_local-header.scss
shaunandrews: Oh my. I somehow had changes on my local that weren't pushed to this PR. Sorry about that. Many of these "in-progress" things have been fixed.

shaunandrews:
> There are a few places with commented-out code
> The padding to the right of the Categories label seems too big
> This should be a single column (#172, which is how it looks on trunk to me, so I don't know what that issue is pointing out)

I must have forgotten to push some changes. This should all be fixed.

> I also noticed that I can hover and interact with the page behind the overlay when the Categories menu is open, which is not the case on trunk.

I just pushed a fix for this.

source/wp-content/themes/wporg-news-2021/sass/components/_local-header.scss
```diff
@@ -133,102 +130,116 @@
...
-						position: fixed;
-						top: 0;
-						left: 0;
-						width: 100%;
-						height: 100vh;
-						background-color: rgba(0, 0, 0, 0.5);
-						z-index: -1;
-						animation-duration: 400ms;
-						animation-name: overlay-fade-in;
+						opacity: 1;
```
ryelle, line 136: I think display would work better here - with `opacity` it's always on screen, so it's overlaying the whole content even when the menu is closed.

source/wp-content/themes/wporg-news-2021/sass/components/_site-header.scss
```diff
@@ -1,33 +1,18 @@
...
+			position: sticky;
+			top: 0;
 
-		/*
-		 * For some reason this element is taller in the offset header than it is in the regular header. That adds
-		 * extra spacing after the header that we don't want.
-		 */
-		display: none;
+			body.admin-bar & {
+				padding-top: var(--wpadmin-bar--height);
```
ryelle: There's actually a core-provided value for this now, so you can use `--wp-admin--admin-bar--height`
```suggestion
				padding-top: var(--wp-admin--admin-bar--height, 0);
```

ryelle (commented):
Looking much better with the rest of your changes 😁  all the things I mentioned before are fixed.

Are the white dots here just part of the brush svg? it seems odd since it kind of lines up with the text, but I don't know if I'd go as far as saying it's a bug.
Screen Shot 2022-01-12 at 1 29 29 PM

The hover/focus colors seem off on the non-blue schemes, I don't know if those were designed.

1. When the text is black, the hover is still light:
    Screen Shot 2022-01-12 at 1 33 08 PM
2. On the black bar, there is no visible hover on "News"
    Screen Shot 2022-01-12 at 1 34 09 PM
3. The little ▪️ is not visible on the black menu (or in the bar ^)
    Screen Shot 2022-01-12 at 1 34 17 PM

ryelle:
Oh also, can you run `yarn workspace wporg-news-2021-theme lint:css --fix` to fix up the linter issues? thanks!

source/wp-content/themes/wporg-news-2021/sass/components/_local-header.scss
shaunandrews: We're able to transition the value of `opacity`, but we can't do that if also has `display: none;`. Are there implication to having it there (but transparent) all the time?

ryelle: Yeah, clicking anywhere on the page opens the menu :)

shaunandrews: lol, really? How did I miss that... I'll look into it.

shaunandrews: My bad. Fixed now. `pointer-events` only work when the menu is open now.

ryelle: 👍🏻  yep, working for me now

shaunandrews:
> Are the white dots here just part of the brush svg? it seems odd since it kind of lines up with the text, but I don't know if I'd go as far as saying it's a bug.

Its part of the SVG mask. I've re-exported the mask with the hope of better reflecting the design.

> The hover/focus colors seem off on the non-blue schemes, I don't know if those were designed.

These should all be resolved now. There are some basics in the Figma (https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=4659%3A8224) but I'm not sure they're 100% correct.

ryelle (approved):
Looks good now — I just pushed a quick change to remove the core-provided margin for the admin bar, which fixes this issue:

Screen Shot 2022-01-12 at 4 41 03 PM

ryelle:
> Its part of the SVG mask. I've re-exported the mask with the hope of better reflecting the design.

Definitely better now, thanks!

shaunandrews:
> remove the core-provided margin for the admin bar

Perfect; I knew there was a way to do it! Thanks.

Issue #58, opened 2021-10-27
Gutenberg: Remove redundant templates/parts


iandunn:
There are cases where we've had to create new templates and template parts that are almost identical to existing ones, because we can't have simple logic inside the HTML files. Examples:

- [ ] `index.html` and `category.html` are identical except for a few class slugs
- [ ] `content-post.html` and `content-page.html` are identical except for the post terms. They could be a single `singular.html` with an `if()` around the terms block.

Once https://github.com/WordPress/gutenberg/issues/32939 is resolved, we should be able to remove a lot of the WET stuff and make things simpler.

Similarly, there's some of duplicated markup:

- [ ] query pagination. we should be able to revert https://github.com/WordPress/wporg-news-2021/pull/179/ once https://github.com/WordPress/gutenberg/issues/28011 is fixed

PR #179, opened 2022-01-11
Move pagination from template-part to the main templates

tellyworth:
It seems that query context isn't passed into template parts(!). This means that query-pagination blocks must be in the same template file as the query block or they won't work.

Related: https://github.com/WordPress/gutenberg/issues/28011

Fixes (most of) #70.


iandunn (approved):
This WFM 👍🏻 

It's unfortunate that we have to make things less DRY, but I added a reminder to fix that once the upstream issue is resolved

https://github.com/WordPress/wporg-news-2021/issues/58

PR #192, opened 2022-01-13
Make the Get WordPress button wider

### LINKED ISSUE #180: `Get WordPress` button should be wide width

danieldudzic:
Closes #180 

News_Theme_2021_Test_–_Testing_github_com_WordPress_wporg-news-2021


iandunn (changes requested):
It's not obvious at first, but the header/footer lives at https://github.com/WordPress/wporg-mu-plugins/tree/trunk/mu-plugins/blocks/global-header-footer, rather than in this theme

So the styles should probably go in https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss

ryelle:
Merged in https://github.com/WordPress/wporg-mu-plugins/pull/84

PR #193, opened 2022-01-13
Footer text and links font size should match the header 13px

ramonjd:
Resolves: https://github.com/WordPress/wporg-news-2021/issues/170

Let's add a navigation style override so we can define the footer nav font size match the default header font size!

| Before  | After |
| ------------- | ------------- |
| Screen Shot 2022-01-13 at 3 52 29 pm  | Screen Shot 2022-01-13 at 3 52 14 pm  |

I'm not familiar with the file system preferences, so I wasn't sure whether to add this style to [sass/components/_site-header.scss](https://github.com/ramonjd/wporg-news-2021/blob/a8e68a045ce487cca7f62956b6d34d8231fcc2c8/source/wp-content/themes/wporg-news-2021/sass/components/_site-header.scss).

Creating a new _navigation.scss file seemed right since we're dealing with a nav block.

The default font-size (`13px`) for the header comes from the header.pcss file generated in `wporg-mu-plugins`.

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/blocks/_navigation.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/blocks/_navigation.scss
@@ -1,0 +1,6 @@
+.global-footer .wp-block-navigation {
+
+	@include break-medium {
+		font-size: 13px;
+	}
+}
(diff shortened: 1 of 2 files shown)


ryelle:
Thanks @ramonjd! Changes to the header & footer should actually be made over in the [WordPress/wporg-mu-plugins](https://github.com/WordPress/wporg-mu-plugins) repo, so I've ported your fix over there: https://github.com/WordPress/wporg-mu-plugins/commit/961a5794ef48789b6f405ace9701d8ae66318d13

ramonjd:
> Changes to the header & footer should actually be made over in the WordPress/wporg-mu-plugins repo, so I've ported your fix over there: WordPress/wporg-mu-plugins@961a579

Thanks @ryelle!

Today I Learned!

PR #195, opened 2022-01-13
Fix local nav bar visual elements

### LINKED ISSUE #194: Local Nav Bar: align visual elements with design
- The separator between `News` and `Category` should be a tiny round dot `·` instead of a squared one
- The word `News` default state color should be blue and have an underline on hover

| Test Site   | Mockup |
| ----------- | ----------- |
| Captura de ecrã 2022-01-13, às 10 21 57 | Captura de ecrã 2022-01-13, às 10 23 05  |

To explain the logic, the only place where there should be a squared dot is to signal the current category page because it's visually similar to the mobile one.

Captura de ecrã 2022-01-13, às 10 30 01
Captura de ecrã 2022-01-13, às 10 31 17

allancole:
This PR changes the square separator to a circular shape and adjusts the default color state for the News link. 

Before: 
image

After: 
image

Figma reference: 
image

Note: I replaced the pseudo element square shape with a unicode "bullet" method which requires a few less lines of code. I tried to simply add a border radius, but once that shape got below 3px, the border-radius had no effect. 

Fixes: #194

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/components/_local-header.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/components/_local-header.scss
@@ -55,9 +55,9 @@
 
 			&::before {
 				content: "\2022";
-				font-size: 0.75rem;
+				font-size: 0.5rem;
 				display: inline-block;
-				margin: 0 12px;
+				margin: 0 0.5rem;
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


ryelle:
@shaunandrews This was changed from a bullet to a square in #178 yesterday, can you take a look here?

shaunandrews:
I reused the CSS I wrote for the current category; I didn't notice it was supposed to be a circle.

> I tried to simply add a border radius, but once that shape got below 3px, the border-radius had no effect.

I don't have a strong preference, but this same method is used in a few other places — using the bullet character feels a little _less_ accurate than actually defining the size in pixels of the pseudo-element.

I just tried it and a border-radius works fine for me — I pushed the changes to make sure.

PR #197, opened 2022-01-14
Add line height value to latest release title

### LINKED ISSUE #165: Latest release line height should be `110px`
As shown [here](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=10002%3A41933)

Captura de ecrã 2022-01-04, às 11 06 23

MaggieCabrera:
This PR adds a line-height value to the latest release title. 

Screenshot 2022-01-14 at 10 35 05

I did not see any grid-related issues coming from this change @apeatling, can you tell me what you saw when you tried this? 

@tellyworth the Latest Release text to the left looks too small compared to the figma files, should we fix this too?

Closes https://github.com/WordPress/wporg-news-2021/issues/165

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/components/front-page/_latest-release.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/components/front-page/_latest-release.scss
@@ -30,6 +30,7 @@ body.news-front-page .front__latest-release {
 			font-size: var(--header-size);
 			font-family: var(--wp--preset--font-family--inter);
 			font-weight: 200;
+			line-height: 1.1;
 		}
 	}
 


ryelle (approved):
Looks good!

ryelle:
> the Latest Release text to the left looks too small compared to the figma files, should we fix this too?

Good catch - want to open another issue for this?

MaggieCabrera:
> > the Latest Release text to the left looks too small compared to the figma files, should we fix this too?
> 
> Good catch - want to open another issue for this?

I think there's a bug where some custom properties have a wrong name, found it with @danieldudzic and I think he's making a PR to fix that, I'm pretty sure it's what's causing that too

apeatling:
I think what I was looking at was the line height of all four sections, which includes the space between "Latest Release" and "See all releases" on the left. The space there also needs to be adjusted.

Issue #148, opened 2022-01-05
Initial Launch: Category Pages


beafialho:
The following category pages are an essential part of the redesign, therefore having them built is critical for the initial launch:

- **Podcast**

| Test Site      | Mockup |
| ----------- | ----------- |
|  <img width="1532" alt="Captura de ecrã 2022-01-05, às 12 04 58" src="https://user-images.githubusercontent.com/65220155/148214899-e2ea686c-e7f3-4f6b-bbcb-8f71147699f8.png">  |  <img width="767" alt="Captura de ecrã 2022-01-05, às 12 05 39" src="https://user-images.githubusercontent.com/65220155/148214964-240fd22f-475e-4b50-b646-e4d560ac5669.png">  |

- **Community**

| Test Site      | Mockup |
| ----------- | ----------- |
|  <img width="1534" alt="Captura de ecrã 2022-01-05, às 12 03 04" src="https://user-images.githubusercontent.com/65220155/148214654-7ecd1146-2d81-4ac3-8690-4461420099fb.png">  |   <img width="670" alt="Captura de ecrã 2022-01-05, às 12 02 39" src="https://user-images.githubusercontent.com/65220155/148214692-ff748128-4103-49ed-80d4-abc3daaae3e7.png">  |

- **Month in WP**

| Test Site      | Mockup |
| ----------- | ----------- |
|  <img width="1479" alt="Captura de ecrã 2022-01-05, às 12 07 29" src="https://user-images.githubusercontent.com/65220155/148215339-43ca496d-82b5-4246-9b77-bff26ae9e4b6.png"> |  <img width="1068" alt="Captura de ecrã 2022-01-05, às 12 08 27" src="https://user-images.githubusercontent.com/65220155/148215352-d2184e50-270a-43fa-a913-730120ce7198.png"> |

- **Releases**

| Test Site      | Mockup |
| ----------- | ----------- |
|  <img width="1527" alt="Captura de ecrã 2022-01-05, às 12 09 58" src="https://user-images.githubusercontent.com/65220155/148215491-5f0715cd-622f-4800-83b9-03bb8c5d2218.png">  |   <img width="971" alt="Captura de ecrã 2022-01-05, às 12 10 31" src="https://user-images.githubusercontent.com/65220155/148215565-6129462c-0a4d-4a88-9a1d-42023593fde7.png"> |

- **Events**

| Test Site      | Mockup |
| ----------- | ----------- |
[...]

PR #198, opened 2022-01-14
Starting styles for month in WP category

MaggieCabrera:
Partially addresses https://github.com/WordPress/wporg-news-2021/issues/148

I went ahead and added a few starting styles to the month in WP category page. I hope it's ok to do this at this stage.

I removed the stroke, I don't think the `last-in-year` class is enough to achieve the design. I'm happy to keep working on that but I wanted to get something that didn't look super broken as a start.

Desktop:

Screenshot 2022-01-14 at 12-01-19 Month in WordPress – wporg-news-2021

Mobile:

Screenshot 2022-01-14 at 12-50-39 Month in WordPress – wporg-news-2021

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
@@ -53,6 +53,9 @@ body.category-month-in-wordpress {
 	.wp-block-post {
 		text-align: center;
 		margin-bottom: 40px;
+		@include break-small-only() {
+			width: 50%;
+		}
 	}
 
 	.wp-block-post-date {
@@ -76,7 +79,7 @@ body.category-month-in-wordpress {
 	}
 
 	h2 time {
-		font-size: 120px;
+		font-size: min(max(50px, 5vw), 120px);
 		line-height: 1;
 		a:hover,
 		a:focus {
--- a/source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
@@ -109,7 +109,8 @@ body.category-month-in-wordpress {
 
 		a:hover,
 		a:focus {
-			text-decoration: none;
+			color: #7b90ff;
+			text-decoration-thickness: 1px;
 		}
 	}
 
--- a/source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
@@ -109,7 +109,7 @@ body.category-month-in-wordpress {
 
 		a:hover,
 		a:focus {
-			color: #7b90ff;
+			color: var(--wp--preset--color--blue-3);
 			text-decoration-thickness: 1px;
 		}
 	}
--- a/source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
@@ -66,16 +66,21 @@ body.category-month-in-wordpress {
 		}
 	}
 
-	.last-in-year::after {
-		content: none;//"-";Removing temporarily
-		display: inline-block;
-		line-height: calc(var(--wp--custom--margin--vertical) * 6);
-		width: 100%;
-		background-image: url(images/brush-stroke-short-blue-4.svg);
-		background-position: center;
-		background-repeat: no-repeat;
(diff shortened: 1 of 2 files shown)


MaggieCabrera:
I went into a rabbit hole until I found a solution for the separator. It's CSS hacky but I'm quite happy with it, I don't think there's a better way of doing this without manipulating the markup in some other way.

This is how it looks:

Screenshot 2022-01-14 at 15-35-22 Month in WordPress – wporg-news-2021

source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
```diff
@@ -45,21 +45,63 @@ body.category-month-in-wordpress {
...
+			}
+		}
+	}
+
+	h2 time {
+		font-size: min(max(50px, 5vw), 120px);
+		line-height: 1;
+		a:hover,
+		a:focus {
+			text-decoration: none;
```
ryelle: This is removing any hover effect - if we don't want an underline, can there be another visual indicator? cc @beafialho

ryelle (changes requested):
This looks good! I imported a few more Month in WordPress posts and the break correctly repeats between years.

There's a minor [accessibility issue](https://developer.mozilla.org/en-US/docs/Web/CSS/display#accessibility_concerns) with using `display:contents`, but as far as i can tell it doesn't affect reading out the content, so I think it's okay here. 

Can you run `yarn workspace wporg-news-2021-theme lint:css --fix` to fix the linter issues?

source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
beafialho: These posts are meant to have an [underline and color change on hover](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=4910%3A25499). Is this what you're referring to?

Captura de ecrã 2022-01-14, às 16 37 14

MaggieCabrera: Fixed now!

MaggieCabrera: Also passed the linting too, thanks for the command!

source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
```diff
@@ -45,21 +45,73 @@ body.category-month-in-wordpress {
...
+		}
+	}
+
+	h2 time {
+		font-size: min(max(50px, 5vw), 120px);
+		line-height: 1;
+
+		a:hover,
+		a:focus {
+			color: #7b90ff;
```
ryelle: is there a custom property for this color?

MaggieCabrera: I couldn't find it, and the header is also using the hex code too for the active color (it's the same)

source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
beafialho: Looks great (I'm able to test locally now!). One question @MaggieCabrera, is the H2 font size used in the month's names `50px`? If it is, could we set this category's titles to be `120px`? Asking because I'm seeing post titles in a smaller size than the intended across the whole site and wanted to make sure that's not intentional.

MaggieCabrera: @beafialho I think what you are seeing is a bug with the names of the variables being incorrect for the font sizes, and it's unrelated to this PR, I think Danny is going to fix that. In this particular case I was using a dynamic value for the font size that goes from 50px on mobile to 120px on desktop, I haven't seen a variable for that, maybe I should create one in case the same is reused somewhere else.

beafialho: I see what you mean, that should be fine then. Thanks!

source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
ryelle: Aha, I just found it- it's `var(--wp--preset--color--blue-3)`

MaggieCabrera: hum, my search didn't find it but you are right!

ryelle (approved):

PR #199, opened 2022-01-14
Start community category styles

### LINKED ISSUE #76: community - final markup and css

MaggieCabrera:
Partially addresses #148

This adds some starting styles to the Community category page

Desktop:

Screenshot 2022-01-14 at 12-43-12 Community – wporg-news-2021

Mobile:

Screenshot 2022-01-14 at 12-43-20 Community – wporg-news-2021

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/block-template-parts/content-category-community.html
+++ b/source/wp-content/themes/wporg-news-2021/block-template-parts/content-category-community.html
@@ -1,10 +1,10 @@
 <!-- wp:group {"tagName":"header","className":"entry-header",style":{"spacing":{"padding":{"top":"30px","right":"30px","bottom":"30px","left":"30px"}}}} -->
 <header class="wp-block-group entry-header">
-	<!-- wp:post-title {"level":3,"isLink":true} /-->
 	<!-- wp:post-featured-image {"isLink":true,"width":"200","height":"200"} /-->
 
 	<!-- wp:group {"className":"entry-meta"} -->
 	<div class="wp-block-group entry-meta">
+		<!-- wp:post-title {"level":3,"isLink":true} /-->
 		<!-- wp:post-date /-->
 	</div>
 	<!-- /wp:group -->
--- a/source/wp-content/themes/wporg-news-2021/block-template-parts/content-category-community.html
+++ b/source/wp-content/themes/wporg-news-2021/block-template-parts/content-category-community.html
@@ -4,7 +4,7 @@
 
 	<!-- wp:group {"className":"entry-meta"} -->
 	<div class="wp-block-group entry-meta">
-		<!-- wp:post-title {"level":3,"isLink":true} /-->
+		<!-- wp:post-title {"level":2,"isLink":true} /-->
 		<!-- wp:post-date /-->
 	</div>
 	<!-- /wp:group -->
--- a/source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
@@ -171,14 +171,10 @@ body.category-community {
 					opacity: 0;
 					color: var(--wp--preset--color--white);
 					position: absolute;
-					top: 0;
 					left: 0;
 					right: 0;
 					bottom: 0;
 					z-index: 1;
-					display: flex;
-					flex-direction: column;
-					justify-content: flex-end;
 				}
 			}
 		}
--- a/source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
@@ -109,7 +109,6 @@ body.category-community {
 		a:hover,
 		a:focus {
 			color: inherit;
-			text-decoration: none;
 		}
 	}
 
--- a/source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
@@ -92,7 +92,12 @@ body.category-community {
 		padding: 0;
 		margin-top: 0;
 
-		@include break-small() {
+		@include break-medium() {
+			padding-left: 0;
+			padding-right: 0;
+		}
+
+		@include break-large() {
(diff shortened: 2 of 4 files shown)


ryelle:
It looks like this template has specific hover/focus styles:
Screen Shot 2022-01-14 at 11 56 06 AM
Screen Shot 2022-01-14 at 11 56 17 AM

MaggieCabrera:
> It looks like this template has specific hover/focus styles: 

Oh I missed those, thank you!

source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
```diff
@@ -88,33 +88,93 @@ body.category-community {
...
+				background-color: var(--wp--preset--color--blue-2);
+				color: var(--wp--preset--color--white);
+				vertical-align: bottom;
+			}
+		}
+		// Featured image only for posts that have one
+		&.has-post-thumbnail .entry-meta {
+
+			@include break-small() {
+				display: none;
```
ryelle: Using `display:none` takes this content off the page for screen reader users too - you should be able to use `@include hide-accessibly;` instead (and then you can use the `show-hidden-accessibly` to unhide on hover/focus).

MaggieCabrera: I used opacity for the hover effects in the end, I think that should be fine for a11y?

beafialho:
@MaggieCabrera it's looking good!

- The black color on hover seems darker in the screenshot. Just wanted to make sure it's set to `#1E1E1E` as intended.
Captura de ecrã 2022-01-17, às 17 24 48

- Could you please make sure that the post titles with images have the same spacing as the ones without images? Those are sitting a little bit lower than the others:
Captura de ecrã 2022-01-17, às 17 28 23

- For tablet sizes, the titles overlap and aren't readable. Is there any chance to improve their responsiveness or is this a Gutenberg issue?

https://user-images.githubusercontent.com/65220155/149815125-af6a7686-5ff5-4f16-a995-39f5b3068346.mov

source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
ryelle: That works! Opacity does not remove the element like display:none does 👍🏻

source/wp-content/themes/wporg-news-2021/block-template-parts/content-category-community.html
```diff
@@ -1,10 +1,10 @@
 <!-- wp:group {"tagName":"header","className":"entry-header",style":{"spacing":{"padding":{"top":"30px","right":"30px","bottom":"30px","left":"30px"}}}} -->
 <header class="wp-block-group entry-header">
-	<!-- wp:post-title {"level":3,"isLink":true} /-->
 	<!-- wp:post-featured-image {"isLink":true,"width":"200","height":"200"} /-->
 
 	<!-- wp:group {"className":"entry-meta"} -->
 	<div class="wp-block-group entry-meta">
+		<!-- wp:post-title {"level":3,"isLink":true} /-->
```
ryelle: The parent heading here is `h1`, so this should be an `h2` to avoid skipping levels ([more info on why here](https://webaim.org/techniques/semanticstructure/#headings)).

Screen Shot 2022-01-17 at 12 44 53 PM

source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
```diff
@@ -88,33 +88,131 @@ body.category-community {
...
+			}
+
+			&:nth-of-type(odd) {
+				background-color: var(--wp--preset--color--blue-2);
+				color: var(--wp--preset--color--white);
+				vertical-align: bottom;
+			}
+		}
+		// Featured image only for posts that have one
+		&.has-post-thumbnail {
```
ryelle: Can this ruleset be merged with the `&.has-post-thumbnail {` above (on line 130)?

source/wp-content/themes/wporg-news-2021/sass/components/_categories.scss
```diff
@@ -88,33 +88,131 @@ body.category-community {
...
+			position: relative;
+
+			@include break-small() {
+				&:hover {
```
ryelle: This shows up correctly on hover, but not when tabbing through the page with the keyboard. You should be able to use `focus-within` to also show the content when the links are focused, ex:

```suggestion
				&:hover,
				&:focus-within {
```

ryelle (changes requested):
This is looking good! I have a few small changes requested, and could you run the linter & fix the issues?

```
```

MaggieCabrera:
@beafialho is it ok if we jump from 4 per column to 3 or 2 when we go to tablet size to avoid the issue you highlighted?

beafialho:
@MaggieCabrera yes, we can use 3 or 2 columns

MaggieCabrera:
ok, this is ready for another review, I fixed all the mentioned issues, and changed the grid to drop to 3, then 2 then 1 column while we lower the viewport width

beafialho:
Looks perfect @MaggieCabrera!

Do you have any idea why this post has the off white background?

Captura de ecrã 2022-01-18, às 09 57 38

MaggieCabrera:
> Looks perfect @MaggieCabrera!
> 
> Do you have any idea why this post has the off white background?
> 

It's marked with the class `.has-post-thumbnail` so it should be showing an image instead of the blue background, I'm not sure if there's a problem with the imported content or what's going on there. I could add the blue background always by default so this is not a problem

MaggieCabrera:
> It's marked with the class `.has-post-thumbnail` so it should be showing an image instead of the blue background, I'm not sure if there's a problem with the imported content or what's going on there. I could add the blue background always by default so this is not a problem

Nevermind that, I did that but the text only appears on hover, as do all the rest of the posts with an image. I think it's best to just make sure that the post has a thumbnail correctly in this case

ryelle:
I have the same post missing a featured image, I think there must've been an issue in the export. If I remove the (broken) featured image, it works as expected.

ryelle (approved):
👍🏻  
Thanks for all your work!

PR #201, opened 2022-01-14
Local Header: Add Top-Border to Dark Variant

shaunandrews:
The "Month in WordPress" category template uses a dark navigation bar, which when paired with the global top bar ends up blending in _too_ much; The current indicator under "News" looks out of place, and generally things feel a little awkward:

image

The design calls for a 1px border, which this PR adds:

image


ryelle (approved):
Looks good 👍🏻

PR #190, opened 2022-01-12
Fix stroke on category pages paginations

MaggieCabrera:
This PR fixes the positioning and color of the stroke on the category pages with pagination.

Screen Capture on 2022-01-12 at 12-54-38

Closes https://github.com/WordPress/wporg-news-2021/issues/185
Closes https://github.com/WordPress/wporg-news-2021/issues/186

PR #202, opened 2022-01-15
Fix the background texture under the pagination for the Releases category

danieldudzic:
Part of #148

Before:
Releases_–_wporg-news-2021

After
after-3


ryelle:
I think this was fixed already as part of #190, is there a bug here I'm not seeing? 

Here's what I see:

| trunk | this PR |
|------|-------|
| Screen Shot 2022-01-17 at 12 58 49 PM | Screen Shot 2022-01-17 at 12 59 18 PM |

shaunandrews:
This is handled over in #239

Issue #115, opened 2021-12-08
Front page: See all… links don't work


ryelle:
The "See all…" links on the front page don't work as expected. They all stay on the front page, loading the next page of the main query, which is just the Latest Posts section. Each link circled should go to a different page

![Screen Shot 2021-12-08 at 17 08 57](https://user-images.githubusercontent.com/541093/145292603-227d57ed-4914-4ac8-80df-a482b4274de8.png)

- "See all releases" should link to the releases category
- "See all posts" should link to an "All posts" page
- "See all people" should probably link to the People of WordPress tag, I don't see a mockup for this (but I'd guess it should be like the community category page)

tellyworth:
See #126 for a partial fix.

"See All Posts" isn't right, I'm trying to think of a good simple SotW hack to get a default post excerpt view with all posts shown.

iandunn:
This looks like a duplicate of #70, but please reopen if there's anything distinct that I missed

> "See all posts" should link to an "All posts" page

That's `/page/2`, background is in https://github.com/WordPress/wporg-news-2021/issues/55#issuecomment-953650339 and these code blocks:

https://github.com/WordPress/wporg-news-2021/blob/b63360fa17be22eb87a3654c68bde9147edfafe3/source/wp-content/themes/wporg-news-2021/functions.php#L136-L171

https://github.com/WordPress/wporg-news-2021/blob/b63360fa17be22eb87a3654c68bde9147edfafe3/source/wp-content/themes/wporg-news-2021/functions.php#L190-L214

ryelle:
That issue doesn't capture these. I'll leave it up to you whether to add these to that issue, reopen this, or create a new issue

- "See all releases" should link to the releases category
- "See all people" should probably link to the People of WordPress tag, I don't see a mockup for this (but I'd guess it should be like the community category page)

> That's `/page/2`

Yep, but it doesn't link to that page.

iandunn:
> "See all releases" should link to the releases category

I was thinking the current `wp:query-pagination-next` should do that, since it's inside a query block that for that category.  Is that not the case?

I know the link is currently broken, but that seems like a bug in our code or upstream, and #70 is for fixing that. I'd prefer to use the proper blocks and dynamic data, rather than hardcoding stuff. I don't mind doing that if we really have to, though.


> "See all people" should probably link to the People of WordPress tag

The posts are pulled from [the Interviews category](https://wordpress.org/news/category/interviews/), so I assumed it'd link there


[...]

PR #203, opened 2022-01-18
Front page: Update links to correct destinations

### LINKED ISSUE #70: Fix broken pagination links

ryelle:
This uses the HTML block instead of pagination, because they're not technically pagination— it shouldn't paginate this query, it should go to a new page. I've left the markup the same (using `wp-block-query-pagination` for the class) to avoid CSS changes for now.

Supersedes #126, Fixes #70.

**To test:**

View the front page, and check the following links:
- The "All Posts" link goes to `/page/2/`, and shows the second page of posts
- The "All Releases" link goes to `/category/releases/` and shows the list of releases
- The "All People" link goes to `/tag/people-of-wordpress/` and shows the list of people

We had a discussion about the "See All People" link destination [here](https://github.com/WordPress/wporg-news-2021/issues/115#issuecomment-993745600), but it wasn't resolved - I've left it linking to the People of WordPress tag, since that's what this query block uses.

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/block-template-parts/front-page/latest-posts.html
+++ b/source/wp-content/themes/wporg-news-2021/block-template-parts/front-page/latest-posts.html
@@ -20,11 +20,11 @@
 		<!-- /wp:group -->
 	<!-- /wp:post-template -->
 
-	<!-- wp:html -->
-	<!-- See https://github.com/WordPress/wporg-news-2021/issues/70 -->
-	<div class="wp-block-query-pagination" style="display:flex;">
-		<a href="/page/2" class="wp-block-query-pagination-next">See All Posts</a>
-	</div>
-	<!-- /wp:html -->
+	
+	<!-- wp:paragraph {"className":"front__next-page"} -->
+	<p class="front__next-page">
+		<a href="/page/2">See All Posts</a>
+	</p>
+	<!-- /wp:paragraph -->
 </section>
 <!-- /wp:query -->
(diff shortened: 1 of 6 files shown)


iandunn (approved):
IMO it'd be better to update the CSS now, but not a big deal.

ryelle:
I'm not sure what these elements should be — if you'll want to change them back to pagination (should gutenberg change how that works), or if they should be links in paragraph tags for editing in the site editor, etc. This was intended to be a fast workaround to fix navigation, not necessarily a long-term solution.

iandunn:
It seems most likely that we'd keep them as links, but that works for me too 👍🏻  

In that case though, I think it'd be best to open a post-launch issue; that way the confusing class is documented, and can be fixed before it gets copied to more themes.

ryelle:
Okay, I've updated it to paragraph + link and a custom class. The main difference now is that the links will always be visible, even if there are no posts in the archive. Given the content on the News site, that should never happen, but it is a change from current behavior.

| Before | After |
|---|---|
| before | after |

iandunn:
That LGTM 👍🏻 

`front__next-page` has duplicate styles, but I can open an issue to make them DRY after the launch.

Issue #39, opened 2021-10-01
Scale font size to maintain optimal characters per line


iandunn:
See https://github.com/WordPress/wporg-news-2021/issues/37#issuecomment-932300306 for background

Figma: https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=8100%3A35705

Implementation: "modular scale", or maybe something simpler like rems? Do a little research; might be enough to check `twenty*` themes.

coreymckrill:
Possibly relevant: https://richtabor.com/fluid-type-scale-theme-json/

iandunn:
some of these examples may also be relevant: https://developer.mozilla.org/en-US/docs/Web/CSS/clamp()

tellyworth:
@kellychoffman @beafialho another issue that you might be able to help with.

kellychoffman:
This feels a bit out of scope for the initial launch - my suggestion is to choose a single font size for all sizes. @beafialho: what do you think?

beafialho:
> my suggestion is to choose a single font size for all sizes. @beafialho: what do you think?

That sounds ok to me.

On a next stage, it can make sense to use an approach similar to https://developer.mozilla.org/en-US/docs/Web/CSS/clamp(), as @iandunn suggested.

iandunn:
@beafialho, can you add mockups for the bottom banner and release template @ `2000px` ? That'd help me avoid guessing on some things.

beafialho:
Sure @iandunn, I'll work on those and place them here.

beafialho:
Here's the bottom banner and release template at `2000px`: https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=11980%3A53269

If others are needed please let me know.

iandunn:
Thanks! It looks like after `1320px`, the left/right edge padding starts growing, in addition to the font size. I'm assuming that should be applied to all pages, but let me know if it's specific to the Releases template.

beafialho:
> Thanks! It looks like after 1320px, the left/right edge padding starts growing, in addition to the font size. I'm assuming that should be applied to all pages, but let me know if it's specific to the Releases template.

[...]

PR #204, opened 2022-01-18
Headings: Increase specificity of responsive font size ruleset

ryelle:
The headings use different font sizes on large screens vs small screens, and to do that, the CSS overrides the custom properties. At some point, either the specificity or the order of the core-generated properties changed, so that they're overriding our overrides… so this change bumps up our specificity, to make sure our values take effect.

| Before | After |
|---|---|
| before | after |

@beafialho Pinging you to make sure these are the correct sizes now :)


beafialho:
Thanks @ryelle! I see the correct sizes on large screens but not on small screens.

ryelle:
Is there something wrong with small screens? I thought the font sizes were already correct there?

beafialho:
> Is there something wrong with small screens? I thought the font sizes were already correct there?

Sorry for the confusion, yes they are correct in the smallest screen size 👍 

I thought they would decrease gradually _as_ we resize the window. Correct me if I'm wrong but I think at some point we were using a function to do that?

ryelle:
> I thought they would decrease gradually as we resize the window. Correct me if I'm wrong but I think at some point we were using a function to do that?

It looks like we're only using that function on `h3`s, and it only applies between 650px - 900px wide.

beafialho:
> It looks like we're only using that function on `h3`s, and it only applies between 650px - 900px wide.

I see, this works fine for now. Could we apply it for other headings on a next phase post launch?

ryelle:
Yeah, I think it's a great idea - looks like there's already an issue for post-launch iteration: https://github.com/WordPress/wporg-news-2021/issues/39

beafialho:
> Yeah, I think it's a great idea - looks like there's already an issue for post-launch iteration: #39

That's right, I kept thinking there was a convo about it somewhere... Thank you!

Issue #75, opened 2021-11-10
Add artistic content to left column


iandunn:
The content here isn't essential for the launch, and isn't straightforward to implement w/ the current state of FSE.

- [x] Releases category: version number the post is about, like described in #38 
- [ ] Events category: year of post. The year is simple with `<!-- wp:post-date {"level":2,"className":"event-year","format":"Y"} /-->`, but it needs to be grouped according to year, which isn't simple. #281 
- [x] All categories without a specific template: The first letter of the category name, with other letters hidden. This could be done in CSS with `color` and `:first-letter`, but that's hacky. [Dynamic templates](https://github.com/WordPress/gutenberg/issues/32939) would probably be best. A custom block would also work, but that isn't a sustainable approach for all the other instances where we need dynamic templates.
- [ ] The 2-column layout styles may need to be adjusted to account for all of the above

ryelle:
Do we need this separate tracking issue? Releases & the initial letters are done, and Events is tracked in #281.

iandunn:
Doesn't seem like it, now that those issues exist 👍🏻

PR #208, opened 2022-01-20
Add dropcap to categories pages

MaggieCabrera:
I'm not sure if this is tracked somewhere but I've seen there's a dropcap in the design of category pages, so I went ahead and implemented it:

Screenshot 2022-01-20 at 11 40 34

Screenshot 2022-01-20 at 11 39 33

Screenshot 2022-01-20 at 11 39 26

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/block-templates/category.html
+++ b/source/wp-content/themes/wporg-news-2021/block-templates/category.html
@@ -2,6 +2,13 @@
 
 <!-- wp:query {"tagName":"main","className":"site-content-container"} -->
 <main class="wp-block-query site-content-container">
+
+	<!-- wp:group {"className":"query-title-banner__title__dropcap"} -->
+	<div class="query-title-banner__title__dropcap">
+		<!-- wp:query-title {"type":"archive"} /-->
+	</div>
+	<!-- /wp:group -->
+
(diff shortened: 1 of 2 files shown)


beafialho:
Thanks @MaggieCabrera, this looks good!

iandunn:
Tracked in #75

iandunn (approved):
#208 has a different approach to this that feels a bit more straightforward, but this one works for me too.

https://github.com/WordPress/wporg-news-2021/blob/3c8e3b5288b8b120ba223a507d5575473ea995a2/source/wp-content/themes/wporg-news-2021/sass/page/_categories.scss#L9-L30

Whichever you prefer works for me, but I think we should make the change in this PR since that one is already doing way too much for a single PR.

iandunn:
~~I do like that this approach is automated, but it may need to be restricted to certain categories.~~

Er, maybe not since those others would have different templates anyway.

MaggieCabrera:
I think this is a better solution than having the letter in the CSS because of i18n. Right now everything that doesn't need the dropcap has a different template like you said. Any exceptions to the rule can also be added easily with an exception to the CSS.

iandunn:
i18n is a good point 👍🏻

PR #210, opened 2022-01-20
Initial layout for podcast single and archive pages

### LINKED ISSUE #130: podcast - final markup and CSS
(design still in flux as of nov 10)

MaggieCabrera:
This PR includes the templates and styles for the podcast archive and single pages. My screenshots show the bug where the heading font sizes are not showing correctly, I couldn't solve this in my local environment using `composer update` but I believe this is not an issue on production.

This PR requires that we uncheck the following on the plugin's settings page so that we don't see duplicates of the player:

image

Archive page: | Single page: 
--- | ---
Screenshot 2022-01-26 at 15-55-47 Podcast – wporg-news-2021 | Screenshot 2022-01-26 at 16-01-19 Episode 14 The Art and Science of Accessibility – wporg-news-2021

The podcast section is not showing up on the menu, you can test this going to http://localhost:8888/podcast/

Closes https://github.com/WordPress/wporg-news-2021/issues/130

Partially addresses https://github.com/WordPress/wporg-news-2021/issues/148


MaggieCabrera:
@beafialho I'm working on this and I'm worried about the font size of the title. It looks like it's designed so the "Podcast" word expands to the width of the container, no matter the viewport. That's tricky when we think about i18n and how that word might look different in other languages. I can only think of doing this using JS so that we can calculate the font size dynamically and the word fits in place, but that's kind of overkill. What do you think about this? What do you expect to happen in languages where the word is bigger than it is in English?

Screenshot 2022-01-21 at 11 32 16

MaggieCabrera:
Also, the podcast plugin looks very different from the design, are we using something else?

Screenshot 2022-01-21 at 11 36 12

beafialho:
Thanks @MaggieCabrera. The word `Podcast` is adopted in the majority of latin languages and also greek. For other alphabets, the term generally doesn't increase tremendously in width with some exceptions.

I've added some examples [here](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=10954%3A42744). An example of a long one I expect would behave similarly to this:

Captura de ecrã 2022-01-21, às 11 05 20

MaggieCabrera:
> Thanks @MaggieCabrera. The word `Podcast` is adopted in the majority of latin languages and also greek. For other alphabets, the term generally doesn't increase tremendously in width with some exceptions.
> 
> I've added some examples [here](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=10954%3A42744). An example of a long one I expect would behave similarly to this:
> 

Yeah, that was what I was afraid of, we'd need javascript to figure out the correct font size depending on the word.

beafialho:
> Yeah, that was what I was afraid of, we'd need javascript to figure out the correct font size depending on the word.

Does anything come to mind to you as a better alternative to that?

MaggieCabrera:
> > Yeah, that was what I was afraid of, we'd need javascript to figure out the correct font size depending on the word.
> 
> Does anything come to mind to you as a better alternative to that?

No, nothing that will work dynamically (looking at https://css-tricks.com/fitting-text-to-a-container/ for ideas, and neither looks great on the JS front to me either). The only option is to find a max font size that we can compromise on.

beafialho:
> The only option is to find a max font size that we can compromise on.

Ok, I suggest setting `150px` as maximum font size. That way, if the word is longer and it jumps to the line below, the blob can increase in height keeping the intended bottom spacing of `120px`. Here's an example below and the mockups for a few others on [Figma](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=10965%3A60006).

Captura de ecrã 2022-01-21, às 15 20 20

iandunn:
> the podcast plugin looks very different from the design, are we using something else?

We have a fair amount of customizations for Seriously Simple Podcasting, so it'd probably be a lot less work if we can match the design w/ custom CSS. If that's not practical, though, we may be able to programmatically swap in another _player_, while keep the _plugin_.

cc @hlashbrooke, who may have some ideas/experience with customizing the player.

hlashbrooke:
Th plugin has two options for a player - one is the built-in WordPress media player that you can style with CSS to some degree (I made [a plugin](https://wordpress.org/plugins/media-player-style-kit/) that allows you to do this from the Customiser, but haven't tested or used it in a good number of years).

The other player is an HTML5 player generated by plugin itself. That was developed after I sold the plugin to its current owners so I don't know too much about it, but @jonathanbossenger will have plenty of knowledge about that since he built it.

jonathanbossenger:
I don't work at Castos anymore, but @hlashbrooke is correct that I implemented the HTML5 player. That player is made up of an [HTML template](https://github.com/CastosHQ/Seriously-Simple-Podcasting/blob/master/templates/players/castos-player.php) and some custom [Javascript](https://github.com/CastosHQ/Seriously-Simple-Podcasting/blob/master/assets/js/castos-player.js) and [CSS](https://github.com/CastosHQ/Seriously-Simple-Podcasting/blob/master/assets/css/castos-player.css).

I don't recall if we made it possible at the time to swop in different HTML and or CSS, and I think for business reasons at the time the decision was made not to make this possible, but let me know if I can be of assistance, either digging into the code to see if it is possible, or submitting a PR to make it possible. (Whether it will be accepted or not is another story, but I can try)

jonathanbossenger:
Update, digging into the code now quickly, it looks like we did make it so that it's possible to swop things out, the CSS is wp_enqueued properly, so that should be possible to dequeue, and the renderer class that renders the HTML template has a filter, so we could swop that out the template content. The implementation is a bit clumsy, but a PR to add a few additional filters should fix that, and given that another filter already exists, I don't see why a new one would be rejected.

iandunn:
Thanks for the info, that's really helpful!

MaggieCabrera:
I don't think this is ready for tomorrow, but it's a solid start. What's left to do, in case anyone wants to pick it up (else I'll continue tomorrow):

- [x] Show player on single page below the header
- [x] Style player following the design in https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=11029%3A70037
- [x] Show player for each post on archive page
- [ ] Make the text of the title in the archive page dynamic/translatable
- [ ] Make the text of the sidebar dynamic/translatable
- [x] Review on mobile

MaggieCabrera:
So the podcast player positioning was a bit of a headache. Thanks, @mikachan for helping me with that because I was going kind of crazy. Let me summarize the problems I've found and the solutions we arrived to:

- The single post for the podcast shows the player in the header area. The plugin includes multiple ways of inserting the player, via shortcode or blocks mainly. The problem with blocks is that they require the episode ID (which is basically the post ID) and we can't access that in the HTML template, so we decided to use the `render_block` filter for that to target a group block with a specific class that will contain the player. There are still a couple of things to be done about the implementation though:

- [x] dequeue the player that is still showing on the post-content area (it's only showing the podcast links, but they have to go)
- [x] use the correct player (the simple one) instead of the HTML5 one.
- [x] style the player colors

- The archive page shows the HTML5 player on each excerpt. Apparently, the plugin supports the player inside the_excerpt but that wasn't working for me at all, so I simply used the same solution that I used on the single post and that worked nicely.

It's not the prettiest way of doing this but we couldn't find a better way, so I'm happy to hear any alternatives. I'm off for today, I think I can finish the rest of the issues tomorrow and leave this one finished finally. Thanks for bearing with me on this one.

mikachan:
I've just pushed a change that handles the simple player as well as the HTML5 one in the filter. I thought I'd test out using the post ID and slug to make up the MP3 URL, and it seems to work fine for the podcasts I tested locally. I'm not sure if this is 100% reliable - but it means we're using the simple player in the single post for now!

I've also changed the podcast plugin settings locally to not show in any location, and this seems to remove the additional player from the post-content area:

image

So now I can see this:

image

If these changes are OK, we then just need to style the simple player.

MaggieCabrera:
> So now I can see this:
> 
> image
> 
> If these changes are OK, we then just need to style the simple player.

Looking at your screenshot now I see how this is the default player for the browser (it looks very different in FF!). Maybe we should still use the big one and style it with CSS? Cause I don't think we can style the native one.

mikachan:
Oh I didn't realise you can't style the native player! Yeah, using the big one and styling that with CSS sounds good then.

MaggieCabrera:
This is ready for review. I've left the text on the templates until we decide how we are going to handle those translations. The archive title for this page shows "Archives: Podcast" instead of WP Briefing Podcast like on the design.

ryelle:
Here's what I see on the current branch:

| 400px | 1020px | 1300px |
|-------|--------|--------|
| archive-400 | archive-1020 | archive-1300 |
| single-400 | single-1020 | single-1300 |

I noticed a few bugs, mostly from testing keyboard navigation. The player buttons have no focus states (both the small player on the single page, and the larger one on the archives). When I tab through the player on archives, I end up being able to tab into the share screen, which kind of scrolls the player out of the container:



Along that line, if I click share or subscribe, the little modal that pops up is hard to read:
share-styles

And lastly, the footer seems to be missing from the archive page:
archive-footer-missing

ryelle:
I forgot to point out that the player looks a little tight on small screens:

Screen Shot 2022-01-26 at 1 32 39 PM

MaggieCabrera:
@beafialho what do you think we should do about the pink blob on tablet-size screens?

MaggieCabrera:
> The player buttons have no focus states (both the small player on the single page, and the larger one on the archives).

What should they look like? Just an outline like on menu elements?

beafialho:
> @beafialho what do you think we should do about the pink blob on tablet-size screens?

@MaggieCabrera I added some tablet-size mockups [here](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=11359%3A107966).

scruffian:
> The player buttons have no focus states (both the small player on the single page, and the larger one on the archives)

I added some CSS for this, so that the focus states match the hover states, but maybe we want something more obvious?

scruffian:
I also fixed the focus states for the share panels:
player

scruffian:
I also added tablet styles for the pink blob:
Screenshot 2022-01-27 at 14 02 04

scruffian:
I think this is ready for another review

ryelle:
For focus styles, the hover is good, but can you also add the dotted outline? That's the focus style we've been using elsewhere on the site (`outline: 1px dotted currentColor` usually works).

Tabbing from the "share" link should bring me to the "Read Post" link, not open up the subscribe & share modals - those should only appear if I activate the link.

https://user-images.githubusercontent.com/541093/151392378-51ff01ad-0f83-4fd1-90b1-05c07fe76c1c.mov

I'm also seeing two players on the archive page now, is that a bug with my local setup or did something change?

Small player looks good on mobile now 👍🏻

scruffian:
> For focus styles, the hover is good, but can you also add the dotted outline? That's the focus style we've been using elsewhere on the site (outline: 1px dotted currentColor usually works).

This works for some of the buttons, but the volume, skip and speed buttons use `mask-image` in the CSS which makes the outline invisible. The only work around for this would be to update the HTML for the player which I am reluctant to do.

> Tabbing from the "share" link should bring me to the "Read Post" link, not open up the subscribe & share modals - those should only appear if I activate the link.

The way this works, the sharing panels are visible by default which is why tabbing goes straight to them. I pushed a change which hides them until they are explicitly opened, but I worry this might break other things.

> I'm also seeing two players on the archive page now, is that a bug with my local setup or did something change?

Fixed

ryelle:
> This works for some of the buttons, but the volume, skip and speed buttons use mask-image in the CSS which makes the outline invisible.

Hmmm, that's too bad. I guess the color change is enough of an indicator if we have that technical limitation, it's not great but we're not trying to meet a WCAG level.

> The way this works, the sharing panels are visible by default which is why tabbing goes straight to them. I pushed a change which hides them until they are explicitly opened, but I worry this might break other things.

Everything seems to work fine in my testing, and tabbing through doesn't open the modals anymore. 👍🏻 

Can you add the focus state to the close button too?
Screen Shot 2022-01-27 at 12 56 19 PM

The small player (on mobile and on the single podcast page) need the focus states too.

scruffian:
> Can you add the focus state to the close button too?

The close button doesn't receive focus because it's not a link or a button :(

ryelle:
> The close button doesn't receive focus because it's not a link or a button :(

Oh, you're right, my mistake. that's too bad 😕

beafialho:
Thanks @scruffian and @ryelle for the adjustments and reviews. Some more notes:

> I also added tablet styles for the pink blob:

This seems different from the mockup. The column of text on the left should increase in width. It would be ideal to have the content on the left always inside the blob. It's also lacking some spacing above the posts.

|   Currently   | Mockup |
| ----------- | ----------- |
|  Captura de ecrã 2022-01-28, às 10 05 02   |    Captura de ecrã 2022-01-28, às 10 09 52 |

I see the players on mobile in the darker blue, they should be in the light blue. The post title shouldn't overlap the pink blob.

| Currently      | Mockup |
| ----------- | ----------- |
|   Captura de ecrã 2022-01-28, às 10 12 30 |  Captura de ecrã 2022-01-28, às 10 12 38  |

In the local nav bar, where it says `Archives: Podcast`, it should just say `Podcast`.

Captura de ecrã 2022-01-27, às 10 37 27

Any idea on why that bar is looking taller than the intended height?

**Current height:**
Captura de ecrã 2022-01-28, às 10 16 38

**Intended height:**
Captura de ecrã 2022-01-28, às 10 18 44

scruffian:
I pushed a change to the small player for focus states:
small player

The problem is that the state of the player changes when its hovered - the buttons are while until you hover the player. I have tried to replicate this in the focus styles but its very subtle. It's not possible to add an outline to these buttons as they use `mask-image`.

scruffian:
Thanks for all the feedback here. I believe I have addressed all of the points raised. Please let me know if there are more issues!

ryelle:
> It's not possible to add an outline to these buttons as they use mask-image.

I wish there was a better solution here… but I understand it's a technical limitation. I'll make a follow-up issue in post-launch iteration to design more consistent focus states & more obvious hover states, I don't think this needs to hold up the full page design anymore.

Can you run the linter before merging?

```
```

Aside from that, from a dev perspective, this is good to go 👍🏻

scruffian:
Thanks @ryelle. I'll wait for a 👍  from @beafialho before I merge...

beafialho:
Thanks a lot @scruffian, it's looking great. Here are a few notes and some things I didn't catch previously.

- These two Subscribe and Share links in the player should use `#3E58E1` as their default color. On hover, they can be underlined.
Captura de ecrã 2022-01-28, às 14 22 51

- The post page isn't showing the category:
Captura de ecrã 2022-01-28, às 14 24 12

- In the post page on mobile, I see two players, can we keep just the one on top?
Captura de ecrã 2022-01-28, às 14 24 40

- On a 908px width, the pink blob is looking perfect. Though at 450px it looks a bit offset.

|   908px    | 450px |
| ----------- | ----------- |
|  Captura de ecrã 2022-01-28, às 17 03 19 | Captura de ecrã 2022-01-28, às 17 03 54  |

Can you make it look positioned in the center?

Captura de ecrã 2022-01-28, às 17 04 40

- And finally, in the archive page let's add more spacing to the top of each post.

| Currently      | Mockup |
| ----------- | ----------- |
|  Captura de ecrã 2022-01-28, às 16 56 05   |  Captura de ecrã 2022-01-28, às 16 58 46 |

scruffian:
> In the post page on mobile, I see two players, can we keep just the one on top?

I don't see this. Is it still happening for you?

scruffian:
> Can you make it look positioned in the center?

I pushed a commit to fix this. How does it look now?

scruffian:
> And finally, in the archive page let's add more spacing to the top of each post.

Done. I think this is ready for another review.

beafialho:
> > In the post page on mobile, I see two players, can we keep just the one on top?
> 
> I don't see this. Is it still happening for you?

Yes, I'm still seeing the two players.

Thank you, all looks better. 

One last adjustment though, 80px was too much, I managed to try out 20px and think that should be enough for the posts spacing 👍 

Captura de ecrã 2022-01-31, às 11 12 06

beafialho (approved):

scruffian:
I merged this. If you're still seeing the double players on the test site lets deal with it in a different PR

PR #233, opened 2022-01-22
Categories archives: Hide the decorative archive title from screen readers

### LINKED ISSUE #211: Category page has two level-1 headings
In #208, a second heading (`h1`) was added to the category page, so this page now has duplicate level-1 headings. There should only be one `h1` per page.

Screen Shot 2022-01-21 at 11 58 30 AM

Since the letter is functionally a decorative background image, I don't think we need to make it available to assistive tech (if we did, we'd probably want to make it verbalize just that letter, to communicate what's visually on the page). It looks like `query-title` must be a heading, so it would be best to take it out of the a11y tree with display: none or visibility: hidden.

cc @MaggieCabrera

ryelle:
This prevents a double level-1 heading from appearing on the page, which can be confusing/incorrect for screen reader users.

Fixes #211

This is a bit of a hack — to hide this content from screen readers but not visually, I need to add the `aria-hidden` attribute to the `div` container. As far as I can tell, there's no way to do this in plain CSS. Adding the attribute to the group block works, but if this template is edited in the site editor it will be wiped out. Since the editor is broken anyway, this is probably okay _for now_, but an alternate solution will be needed.

Screenshots:
There should be no visual change:

cat-m

cat-s

I also tried adding a category in Korean to see how this will handle non-Latin characters, and it's okay— the character breaks out of the container a bit - so if/when this theme is used on rosetta sites we'll need to address that (i'm sure it won't be the only thing). I am impressed that the CSS `first-letter` algorithm understands what to do here, and if we tried to implement our own solution in PHP (my only other idea), we'd need to be just as thorough.
cat-han

Heading structure

```
1. Security
  2. WordPress 5.7.2 Security Release
  2. WordPress 5.7.1 Security and Maintenance Release
  2. WordPress 5.4.2 Security and Maintenance Release
  2. WordPress 5.4.1
  2. Get the Latest Updates
    3. Follow The Code
    3. Find An Event Near You
    3. Subscribe to WordPress News
    3. WP Briefing — The WordPress Podcast
```

To test:

- View a category page
- At a screen size >1280px wide, you should see the first letter of the category
- Use a screen reader or other tool to get a list of the headings on the page
- There should only be one `h1` heading

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/block-templates/category.html
+++ b/source/wp-content/themes/wporg-news-2021/block-templates/category.html
@@ -4,7 +4,7 @@
 <main class="wp-block-query site-content-container">
 
 	<!-- wp:group {"className":"query-title-banner__title__dropcap"} -->
-	<div class="query-title-banner__title__dropcap" aria-hidden>
+	<div class="query-title-banner__title__dropcap" aria-hidden="true">
 		<!-- wp:query-title {"type":"archive"} /-->
 	</div>
 	<!-- /wp:group -->
--- a/source/wp-content/themes/wporg-news-2021/block-templates/category.html
+++ b/source/wp-content/themes/wporg-news-2021/block-templates/category.html
@@ -4,7 +4,7 @@
 <main class="wp-block-query site-content-container">
 
 	<!-- wp:group {"className":"query-title-banner__title__dropcap"} -->
-	<div class="query-title-banner__title__dropcap">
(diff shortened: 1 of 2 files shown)


shaunandrews:
I can confirm this fixes the duplicate title when using VoiceOver.

This could be a pretty cool block.

shaunandrews (approved):

Issue #229, opened 2022-01-22
alignfull images not always filling screen


iandunn:
https://wordpress.org/news-test/2021/12/08/people-of-wordpress-devin-maeztri/

none of the ones there are, but i have seen others working

might be #207 regression

PR #235, opened 2022-01-24
Alignments: Add negative margins to .alignfull

### LINKED ISSUE #231: alignfull containers act like alignwide
the container (green background) should stretch to the edges

Screen Shot 2022-01-22 at 9 28 07 AM

#207 regression

mikachan:
This is a potential fix for #231. I've added negative margins to `alignfull` which have the same width as the left and right padding for `entry-content`. This means anything that is set to `alignfull` should stretch to the edges:

image

Is this the intended style for `alignfull`?

Closes #231


mikachan:
This may also fix #229.

shaunandrews:
> Is this the intended style for alignfull?

I believe so. This PR seems to fix #231 and #229 Thanks!

PR #239, opened 2022-01-24
Update/pagination control

### LINKED ISSUE #238: Pagination brush stroke matches link color
image

We need to change the color of the brush stroke to the dark blue.

shaunandrews:
Adjusts the color of text and the brush stroke to work on the blue background of the Releases section (#238):

Before:
image

After:
image

--

Shows the dot between "Newer Posts" and "1 2 3" only when needed, and adjust the overall spacing around the dots:

Before:
image

image

After:

image

image

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/blocks/_query-pagination.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/blocks/_query-pagination.scss
@@ -63,6 +63,7 @@
 			padding: 0 8px;
 		}
 	}
+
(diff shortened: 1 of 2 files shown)


ryelle (changes requested):
Looks good! Just one issue, the last page of posts still has the final dot (ex, on page 4 here):

Screen Shot 2022-01-24 at 4 04 56 PM

shaunandrews:
Pushed a fix that should fix that:

image

image

image

ryelle (approved):

PR #241, opened 2022-01-25
Fix/code blocks

### LINKED ISSUE #161: Posts: Code block
Code block appearance should match the [mockups](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=4945%3A31007):

| Test Site   | Mockup |
| ----------- | ----------- |
|  Captura de ecrã 2022-01-07, às 11 58 10 |   Captura de ecrã 2022-01-07, às 11 56 50  |

tellyworth:
Should fix #161.

To test locally, make sure you run this in addition to the usual build/start instructions:

`composer update && wp-env start --update`

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/blocks/_code.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/blocks/_code.scss
@@ -1,4 +1,5 @@
 // TODO: This can be removed when Gutenberg applies fontFamily correcly to .wp-block-code code (https://github.com/WordPress/gutenberg/issues/31135)
 .wp-block-code code {
 	font-family: var(--wp--custom--code--typography--font-family);
+	border: none;
 }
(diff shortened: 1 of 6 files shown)


source/wp-content/themes/wporg-news-2021/sass/blocks/_code.scss
```diff
@@ -1,4 +1,5 @@
 // TODO: This can be removed when Gutenberg applies fontFamily correcly to .wp-block-code code (https://github.com/WordPress/gutenberg/issues/31135)
 .wp-block-code code {
 	font-family: var(--wp--custom--code--typography--font-family);
```
tellyworth: Can we get rid of the Gutenbug workaround now?

iandunn: That was probably an artifact from forking Blockbase, so it's probably safe to remove if you don't see any obvious side-effects

ryelle: Technically this isn't doing anything because `--wp--custom--code--typography--font-family` is `monospace`, which is the default for `<code>`… but I think there's another (unrelated) bug here, because I don't see any of the block styles from `theme.json` applying.

ryelle: Found & fixed the issue I mentioned: https://github.com/WordPress/wporg-news-2021/commit/5f2db9d753bed8531a13b637522981465895b84f.

PR #242, opened 2022-01-25
Fix/drop cap h1

### LINKED ISSUE #211: Category page has two level-1 headings

tellyworth:
A quick stab at fixing #211.

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/block-templates/category.html
+++ b/source/wp-content/themes/wporg-news-2021/block-templates/category.html
@@ -5,7 +5,7 @@
 
 	<!-- wp:group {"className":"query-title-banner__title__dropcap"} -->
 	<div class="query-title-banner__title__dropcap">
-		<!-- wp:query-title {"type":"archive"} /-->
+		<!-- wp:query-title {"type":"archive","level":"2"} /-->
 	</div>
 	<!-- /wp:group -->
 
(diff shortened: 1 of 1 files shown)


ryelle:
@tellyworth I fixed this in #233, I was just hoping for a review before merging it because of the `aria-hidden` workaround. This fix (switching it to a level-2 heading) is less bad, but still not a great solution since it doesn't really correspond to the page structure.

Edit: I see you also ended up with `aria-hidden`, so I'll consider that a stamp-of-approval 😁

shaunandrews:
I approved #233 with the `aria-hidden`.

I think the issue here isn't so much the heading level — H1 or H2 — but that this uses a heading tag at all. It could be a cool decorative Dropcap block.

ryelle:
> I think the issue here isn't so much the heading level — H1 or H2 — but that this uses a heading tag at all. 

Yep, exactly.

PR #243, opened 2022-01-25
Use a filter to reorder and trim the category list for the menu

### LINKED ISSUE #146: Local Nav Bar: Rearrange Categories
In the test site, this menu is arranged differently than in both the site and the mockup. Here, it seems to be organized alphabetically.

Captura de ecrã 2022-01-04, às 15 40 56

I suggest we organize items by relevance (in the mockup the number of posts in each category was taken in consideration). In the case of "Podcast" we can place it right after "Events", since it's a category we want to highlight.

Captura de ecrã 2022-01-04, às 16 23 32

I believe this is critical for the initial launch, since it can create confusion for users navigating Categories having them be rearranged often, but I look forward to hear your thoughts.

cc: @kellychoffman

tellyworth:
Use a PHP filter to sort the list of categories for the menu, and exclude little-used cats.

This is a bit crude and might need tweaking. Could it break something?

Fixes #146.

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/functions.php
+++ b/source/wp-content/themes/wporg-news-2021/functions.php
@@ -22,6 +22,7 @@ add_filter( 'theme_file_path', __NAMESPACE__ . '\conditional_template_part', 10,
 add_filter( 'render_block_data', __NAMESPACE__ . '\custom_query_block_attributes' );
 add_filter( 'template_redirect', __NAMESPACE__ . '\jetpack_likes_workaround' );
 add_filter( 'the_title', __NAMESPACE__ . '\update_the_title', 10, 2 );
+add_filter( 'get_terms', __NAMESPACE__ . '\filter_categories_for_menu', 10, 4 );
 
 /**
  * Register theme support.
@@ -388,4 +389,34 @@ function update_the_title( $title, $id ) {
 	}
 
 	return $title;
+}
+
+/**
+ * Filter and re-order categories for the menu.
+ *
+ * @param array $terms	The found terms.
+ * @param array $taxonomies	The taxonomies.
+ * @param array $args The query arguments.
+ * @return array The filtered terms.
+ */
+function filter_categories_for_menu( $terms, $taxonomies, $args ) {
+	if ( !is_admin() && ['category'] === $taxonomies ) {
+
+		// First sort the categories by post count descending.
+		usort( $terms, function ( $a, $b ) {
+			return $b->count <=> $a->count;
+		} );
+
+		// Now hide anything below a certain threshold of post counts, or that we explicitly don't want.
+		$hide_slugs = [
+			'uncategorized'
+		];
+		foreach ( $terms as $i => $term ) {
+			if ( $term->count < 5 || in_array( $term->slug, $hide_slugs, true ) ) {
+				unset( $terms[ $i ] );
+			}
+		}
+	}
+
+	return $terms;
 }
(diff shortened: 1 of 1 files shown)


dd32:
Is it possible to set the required args via the `get_term_args` filter at all? That's the direction i glanced at previously

shaunandrews:
Here's what I see locally, but I'd expect this to look different on the live site as it has many more posts for each category.

image

It seems this PR sorts by post count. I wonder if that's the best way to determine what to show — or if it should be a more curated list of categories?

iandunn (commented):
It seems like we need a final decision from Design about what they want, before we can move forward. @beafialho / @shaunandrews , can you work that out and let us know?

shaunandrews:
If @beafialho or anyone else gave specific directives, then by all means listen to them. I was simply trying to understand the effects this PR would have on the live site and if post count is the best way to curate this list. My intention wasn't to put up a roadblock towards progress.

tellyworth:
It's showing fewer category items on a local install because it uses a post count threshold (at least 5 posts) and the local env only imports a limited number of posts.

Here's what I get on a sandbox of the live site using post count ordering and the 5 post threshold:
Screen Shot 2022-01-26 at 2 06 02 pm

tellyworth:
And here's what I get using `get_term_args` filtering, orderby count desc and a limit of 20:
Screen Shot 2022-01-26 at 2 38 35 pm

Note that the menu displays the items sorted horizontally (left-to-right then top-to-bottom). My eyes want them to be sorted vertically. `Meta` and `Events` are reversed between this and the previous screenshot because they have the exact same post count and the different sort algos (php vs mysql) handle that differently.

I think this is probably ok as a starting point.

I thought about using `'orderby' = 'include'` in the `get_terms` args, coupled with an `include` list, which would let us specify the categories and orders directly. But that means using term IDs, which don't translate between the local test and live sites, so I'd resist that solution.

I think probably we should go with something like this for launch, and if/when a more curated list is required replace it with a menu block so it can be user-edited.

I'll update the PR with a `get_term_args` version for review.

beafialho:
When we were sorting out the categories for the design, we assumed these were the only ones so the proposal was based on them.

Captura de ecrã 2022-01-26, às 10 36 47

However, as shown in this [screenshot](https://github.com/WordPress/wporg-news-2021/issues/146#issuecomment-1018165943), there are many more categories, plenty of them with two or three posts. The list is too long and some of those could be deleted and merged with others or simply hidden from the dropdown. For instance the two posts under "WrapUp" could easily fall in their primary categories. An example of what I mean:

| Screenshot      | Rearrangement |
| ----------- | ----------- |
|  150464964-a5a13889-c15b-4320-99e2-11886f5e71e2   |  Captura de ecrã 2022-01-26, às 12 06 41 |

> Without just deleting the old categories, I think the cleanest option here is to add some more clarification here - specifically, what should the category dropdown show?

If we do not want to delete categories, and to answer @dd32's question in #146, here's how I think the dropdown should look. 

- They're sorted alphabetically and vertically
- I've thought about **Design** being merged with **User Interface**, as it's a broader term
- **Heropress** merged with **Community**, also a broader term
- I've added **Store** and **Awards**, as they seem specific enough terms to keep

Captura de ecrã 2022-01-26, às 12 01 13

Looking forward to know if this is achievable.

ryelle:
@tellyworth or @kellychoffman, Do you know if we can edit the categories on News? I like @beafialho's ideas about merging & deleting some categories, then I think we don't need to filter anything (except to add the All posts link).

josephahaden:
I have no problem with this as long as no technical hurdles/blockers exist.

tellyworth:
We'll still need to alter the sort order in CSS from row-column to column-row.

Does it make sense to have a limit on either the category count and/or menu size, as a simple defensive measure?

Likewise for excluding Uncategorized - I know there's the intention to remove posts from that category but it could creep back in future.

ryelle:
> Likewise for excluding Uncategorized - I know there's the intention to remove posts from that category but it could creep back in future.

"Uncategorized" was renamed to "General" in #256, and General is in the list 👍🏻

tellyworth:
Ah got it. So really we need to ditch the PHP and replace this PR with one that reorders the sort from horizontal to vertical.

tellyworth:
Closing this in favour of #277.

PR #119, opened 2021-12-09
Latest Release: Update style to match design

ryelle:
Fixes #114 — Updated the font, spacing, and removed "WordPress" from this post title.

Screen Shot 2021-12-09 at 4 49 55 PM
Screen Shot 2021-12-09 at 4 48 19 PM

Issue #38, opened 2021-09-30
Implement design for Releases category


coreymckrill:
![releases](https://user-images.githubusercontent.com/916023/135541557-4d4f641b-04c4-4f19-886f-45f0cfa26472.jpg)

* [Design](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=4852%3A36290)
* [Interactions](https://video.wordpress.com/v/Da55hR9D)

I'm not sure this one will be possible using block templates. The majority of the posts in this category don't have any meta data (taxonomy term, etc.) to identify which version it relates to; just the version number as part of the post title. So we'll either need to add a taxonomy or post meta value for the version, or write some regex to try and scrape it from the post title. Either way, we'll probably need a PHP template in order to output the markup accurately.

And then there are a few posts that aren't actually about a WP release (like [this one](https://wordpress.org/news/2021/02/reflecting-on-gutenbergs-100th-release/)). Not sure how to present those...

- [x] rough markup #51
- [x] final markup and CSS

iandunn:
🤔 

I'm generally leery of using regex -- _("...now you have two problems")_ -- but I also think it'd be impractical to retroactively apply terms for a new taxonomy, and also impractical to expect authors to use the taxonomy in the future.

I wonder if a really simple regex like `\d(\.\d)+` would be good enough?

https://regex101.com/r/iKnTvJ/1

If that doesn't match anything, then a second regex could search for `{number}{ordinal}` ( _psuedoregex_ 😉 )

...and if that fails, then maybe we leave the space blank? It's probably not worth the time/tech debt to get the last 5% of them.

iandunn:
That regex could be used to retroactively apply a taxonomy term, but we'd still have the problem of folks having to know that they need to use it.

iandunn:
Another idea would be to add that "version number" area part of post content, so it's obvious to folks that they need to add it? Placeholder text could give instructions. 

But then it'd have to be hidden on the single post view. Maybe we could add it in the editor as an inline panel, so that the editor would appear more like the category view. but instead of saving to `post_content`, it'd  save as a taxonomy term. That seems like way too much trouble for something like this, though.

coreymckrill:
Here's a slightly more complex expression `(\d{0,3}(\.\d{1,3})+)` that gets all but 12 of the post titles in the archive:

https://regex101.com/r/NgC94E/1

[...]

PR #246, opened 2022-01-25
"Releases" Template

shaunandrews:
image

This PR updates the Releases category template to more closely match the design. It also includes a nifty,  new block for displaying a release version number.


iandunn:
What do you think about a custom block that pulls the `x.y` out of the title w/ a regex? #38 has the regex worked out already. That seems like less technical debt to me, because it's more straightforward and future-proof. 

We could also use it on the homepage to match the design closer, and remove this (see #119):

https://github.com/WordPress/wporg-news-2021/blob/d55bf56f7e8973660e430b8d0fabc42dec24829a/source/wp-content/themes/wporg-news-2021/functions.php#L376-L391

Something similar would work for the date in the Events template.

Related: #75

tellyworth:
I pushed a commit to add a custom block. Seems to work ok. @shaunandrews I've only lightly touched your CSS, the rest of it is additions.

To get it to work you'll probably need to:

`composer update && wp-env start --update`

@iandunn have I added the block plugin in a sane way? I copied the gitignore handling of the theme folder basically.

.gitignore
```diff
@@ -30,5 +31,9 @@ README-svn.txt
 !/source/wp-content/themes
 !/source/wp-content/themes/wporg-news-2021
 
+# And this plugin.
+!/source/wp-content/plugins
+!/source/wp-content/plugins/wporg-post-title
```
iandunn: 🤔 , I'd lean towards putting the code in the `blocks/` folder in the `mu-plugins` repo, so that all sites can use it. I'd guess that future redesigns will have a similar concept, with part of the post title being used as a design element.

I don't feel strongly, though.

source/wp-content/plugins/wporg-post-title/wporg-post-title.php
```diff
@@ -0,0 +1,75 @@
...
+ * Renders the `wporg/post-title` block on the server.
+ * Copied from core post-title.php and modified.
+ *
+ * @param array    $attributes Block attributes.
+ * @param string   $content    Block default content.
+ * @param WP_Block $block      Block instance.
+ *
+ * @return string Returns the filtered post title for the current post wrapped inside "h1" tags.
+ */
+function render_block_wporg_post_title( $attributes, $content, $block ) {
```
iandunn: I was imagining something much more simple, that would only be used for this purpose, since `core/post-title` can be used if they want the full title.

```php
register_block_type(
	'wporg/post-title-version',
	array(
		'title'           => 'Post Title Version',
		'render_callback' => function() {
			$full_title = get_the_title();
			preg_match( '/WordPress (\d+[.]\d+)/', $full_title, $matches );
			
			return $matches[1] ?? ''; 
		}
	)
);
```

That's probably missing a few things, but communicates the jist of it. I don't think we need `block.json` etc for a small workaround like this. In my mind this is just a temporary thing until  Gutenberg solves https://github.com/WordPress/gutenberg/issues/32939.

source/wp-content/plugins/wporg-post-title/block.json
```diff
@@ -0,0 +1,49 @@
+{
+	"apiVersion": 2,
+	"name": "wporg/post-title",
+	"title": "WordPress.org Post Title",
+	"category": "theme",
+	"description": "Extends the core Post Title block with additional attributes needed for WordPress.org themes.",
```
ryelle: Aside from the `releaseVersion` toggle, how is this different from the Post Title block? Only asking because when I saw "custom block for the version" I thought of making a simple block that _only_ outputs the release version, with no other attributes.

.gitignore
ryelle: I don't think other sites will need this exact functionality, but if we leave it here that's a third folder we would need to sync into SVN.

source/wp-content/plugins/wporg-post-title/block.json
tellyworth: Yeah I like that idea. I guess I was preempting other hypothetical uses for the same block, but I don't think we have any other actual use cases right now.

.gitignore
tellyworth: If I change it to a more specific block like post-title-version, would any other sites have a use for it? It's a pretty obscure use case. Agree that mu-plugins could make sense for a more generally reusable block though.

iandunn: Kelly has a good point. I could even seen this being put in `functions.php` since it's really just a glorified custom template tag. I wouldn't do that for normal blocks, but this is just a workaround for FSE limitations. I don't feel strongly though.

ryelle: I'm okay with leaving it in the theme (maybe a `blocks` folder, in case we need any other workarounds). If we're still going the parent-theme route, they'll be available to any child themes too. I do think we should use a `block.json` + set it up in JS too, so that it won't break in the site editor (see https://github.com/WordPress/wporg-mu-plugins/pull/26) when we get there.

tellyworth:
I changed the block to `wporg/release-version`. It just outputs the version number and a wrapper tag, rather than duplicating post title stuff.

It feels like overkill. The alternative to adding the block in a plugin is to register a block from the theme itself, which I think(?) is generally discouraged.

ryelle:
> The alternative to adding the block in a plugin is to register a block from the theme itself, which I think(?) is generally discouraged.

That's true, but that actually was [what I was suggesting](https://github.com/WordPress/wporg-news-2021/pull/246#discussion_r793151771) 😅  I guess it depends on whether you think we'll need this on sites that don't use this theme (itself or as a parent), or if we might want it in the future, whenever this theme is eventually retired. My opinion is that it's decorative in a way that's tied to the theme, so I think keeping it in the theme is fine.

shaunandrews:
My poor attempt to rebase seems to have broken this PR. Sorry. Trying (and currently failing) to get it sorted out.

source/wp-content/plugins/wporg-release-version/wporg-release-version.php
```diff
@@ -0,0 +1,65 @@
...
+	$align_class_name = empty( $attributes['textAlign'] ) ? '' : "has-text-align-{$attributes['textAlign']}";
+
+	if ( isset( $attributes['level'] ) ) {
+		$tag_name = 0 === $attributes['level'] ? 'p' : 'h' . $attributes['level'];
+	}
+
+	$version = '';
+	$title = get_the_title( $post_ID );
+	// Do we also want x.y.z?
+	if ( preg_match( '/WordPress (\d+[.]\d+)/', $title, $matches ) ) {
```
coreymckrill: A regex that will work for getting 3rd-level versions is `(\d{0,3}(\.\d{1,3})+)` (from [here](https://github.com/WordPress/wporg-news-2021/issues/38#issuecomment-931804064))

iandunn: 👍🏻  It seems important to get `.z` in there

shaunandrews:
Right now the "5.8" version number is duplicated many times, since each of the latest posts (from the import) is for 5.8 release candidates. Would it make sense to update these version numbers to include "RCx" to help reduce the repetition? This might require some design thinking, as I think "5.8 RC2" might be _too long_ to fit within the available space.

beafialho:
Not sure why I'm not seeing the numbers on the left?

Captura de ecrã 2022-02-01, às 11 35 10

tellyworth:
> Not sure why I'm not seeing the numbers on the left?

Did you run this? -

`composer update && wp-env start --update`

tellyworth:
I updated the PR to:

* Remove the separate plugin, and register the block from `functions.php`
* Handle `x.y.z` and `RC n`

Do we also want it to understand `Beta n`? That text is probably too long to fit.

Screenshot (zoomed out to fit more examples):
Screen Shot 2022-02-02 at 12 20 47 pm

Make sure you run `composer update && wp-env start --update` to test it locally.

beafialho:
> Did you run this? -

I hadn't, that worked. Thanks!

> Do we also want it to understand Beta n? That text is probably too long to fit.

Let's keep the numbers only as they fit just fine. I like the "RC" alternative.

Overall, this looks great! I have some comments:

**Desktop**

- The height of each post should ideally be `174px` instead of the current `232px`:

| Currently      | Mockups |
| ----------- | ----------- |
|   Captura de ecrã 2022-02-02, às 14 38 09  |     Captura de ecrã 2022-02-02, às 14 40 13 |

- Increase spacing down at the bottom of the page (should look the same as on top of posts navigation):

| Currently      | Intended |
| ----------- | ----------- |
|   Captura de ecrã 2022-02-02, às 14 30 23  |  Captura de ecrã 2022-02-02, às 14 31 06 |

- Hover effect missing in Read Post (underline should disappear on hover):

https://user-images.githubusercontent.com/65220155/152176296-c992d367-ab7e-492d-98bd-68c4e3db6c5d.mp4

**Mobile**

- Let's remove the arrow on mobile:

Captura de ecrã 2022-02-02, às 14 34 37

- Increase spacing between posts (top and bottom spacing should be identical)

| Currently      | Intended |
| ----------- | ----------- |
|   Captura de ecrã 2022-02-02, às 14 34 50  |  Captura de ecrã 2022-02-02, às 14 36 31 |

- Posts navigation off the page:

https://user-images.githubusercontent.com/65220155/152177013-ea38882c-5790-4b86-8aef-ff48042c3595.mp4

ryelle:
@coreymckrill I made a few changes to the block markup, since we don't _need_ to have this in a heading we can avoid the aria workarounds. The content should still be available to screen reader users (just not as a heading). Feel free to undo anything you don't like :)

I think all of @beafialho's [feedback](https://github.com/WordPress/wporg-news-2021/pull/246#issuecomment-1028018398) is still relevant - I removed the arrow on small screens while restructuring some of the CSS, but things like the post height & wonky pagination on small screens are still happening.

coreymckrill:
@beafialho @ryelle Sorry I missed that some of the feedback from earlier hadn't been addressed yet. It has been now, would love to get your review. One thing to note is that I also moved the version numbers for the not-latest release posts down to the bottom edge of the rows on mobile to better match the Figma design.

beafialho (approved):

beafialho:
Thank you @coreymckrill this looks great!

source/wp-content/themes/wporg-news-2021/sass/page/_categories.scss
```diff
@@ -14,31 +14,174 @@ body.category-events {
...
+		border-bottom: 1px solid var(--wp--preset--color--blue-5);
+
+		// This is needed to position the version number so
+		// it appears "hidden".
+		overflow: hidden;
+
+		// Ignore some of this extraneous markup.
+		.wp-block-template-part {
+			display: contents;
+		}
```
ryelle: `display: contents` can have [some accessibility issues](https://developer.mozilla.org/en-US/docs/Web/CSS/display#display_contents), and I don't understand why it's needed here - can you explain a little more?

source/wp-content/themes/wporg-news-2021/sass/page/_categories.scss
```diff
@@ -14,31 +14,174 @@ body.category-events {
...
+		.wp-block-group.entry-header,
+		.wp-block-post-title,
+		.wp-block-post-date {
+			margin: 0;
+		}
+
+		.wp-block-post-title {
+			// I wanted to use a var here, but there isn't one
+			// for this value.
+			font-size: 36px;
```
ryelle: I think `var(--wp--custom--h-3--typography--font-size)` would work, it's 26px on small screens but grows to 36px on larger.

source/wp-content/themes/wporg-news-2021/sass/page/_categories.scss
```diff
@@ -14,31 +14,174 @@ body.category-events {
...
+				}
+			}
+
+			.wp-block-post-excerpt {
+				display: block;
+			}
+
+			.wp-block-post-excerpt__more-link {
+				&:hover,
+				:focus {
```
ryelle:
```suggestion
				&:focus {
```

ryelle (commented):
Overall this is working great, I just caught a few things while looking through the code.

I also noticed the pagination is missing space between 600-780px:

Screen Shot 2022-02-14 at 11 03 49

source/wp-content/themes/wporg-news-2021/sass/page/_categories.scss
coreymckrill: 🤔 That was added [before I got involved](https://github.com/WordPress/wporg-news-2021/commit/cd77182f6bd57cc1552ac37e3b601155dd3b547b), so I'm not sure why it's there. @shaunandrews do you remember why it's there?

FWIW removing it doesn't seem to affect anything.

source/wp-content/themes/wporg-news-2021/sass/page/_categories.scss
coreymckrill: Also before my time, I'm fine with changing it: 012ece4

source/wp-content/themes/wporg-news-2021/sass/page/_categories.scss
coreymckrill: I went ahead and removed it in 932a4f7, but if it turns out it's needed for something, we can discuss putting it back in.

Issue #223, opened 2022-01-19
Pages with short content have gap under footer


iandunn:
- [x] fix content pushed to middle in this theme - https://wordpress.org/news-test/?p=11280&preview=true

fix gap under footer in classic themes:
- [ ] https://translate.wordpress.org/projects/
- [ ] wporg-main
- [ ] check other themes

older themes do that too, so it's not a regression there, but would be nice to fix. we can push those to post-launch or try to do it before

iandunn:
Actually, this is showing up w/ the block theme too: https://wordpress.org/news-test/?p=11280&preview=true

There was originally some CSS in the this repo or the theme which prevented that, but it may have been broken by another change.

dd32:
I was unable to reproduce this using either of these, perhaps because It's not actually clear as to what the problem is to me.
- super short: https://wordpress.org/news-test/?p=11280&preview=true
- super long: https://wordpress.org/news-test/?p=11295&preview=true

iandunn:
There's white space under the footer instead of the footer being stuck to the bottom:

<img width="524" alt="Screen Shot 2022-01-25 at 11 47 54 AM" src="https://user-images.githubusercontent.com/484068/151048576-16b919f0-2ff9-4e69-b996-b0ceee1c4562.png">

_( you may need to add a `border` to that `img` in dev tools to make it obvious, or open the URL and set responsive mode to a 2000px tall viewport )_

I've seen it on https://wordpress.org/news-test/?p=11280&preview=true too at some breakpoints. I think there are probably some examples like that on most sites, but fixing it will vary somewhat by the theme

iandunn:
In the News theme, it does stick to the bottom, but the way it does has creates a side effect where the post content floats down to the middle, instead of sticking to the top: 

<img width="1354" alt="Screen Shot 2022-01-25 at 1 36 33 PM" src="https://user-images.githubusercontent.com/484068/151063796-f8f64708-583e-40b6-bce3-2226054db213.png">

iandunn:
This fixes translate.w.org pages:

```diff
diff --git templates/style.css templates/style.css
index da61356..2840162 100644
--- templates/style.css
+++ templates/style.css
@@ -9,8 +9,11 @@
        }
 }

-body, html {
-       height: 100%;
+body {
+       /* This allows the bottom margin on `gp-content` to push the footer to the bottom of short pages. */
+       display: flex;
+       flex-direction: column;
[...]

PR #247, opened 2022-01-25
Layout: Stick content to top of page.

iandunn:
See #223

The previous method of sticking the footer to the bottom had the side-effect of pushing the content down to the middle. This achieves the former without the latter.

<table>
<tr>
<td>before:

Screen Shot 2022-01-25 at 14 30 07-fullpage

</td>
<td>after:

Screen Shot 2022-01-25 at 14 29 56-fullpage

</td>
</tr>

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/base/_layout.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/base/_layout.scss
@@ -6,9 +6,11 @@
 	flex-direction: column;
 	margin-top: -(var(--wp-admin--admin-bar--height, 0));
 
-	// Make sure the footer is always at the bottom of the viewport, even on pages with little content.
-	// Otherwise there'd be empty space below the footer
-	justify-content: space-between;
+	> .site-content-container {
+		// Make the content area grow to fill any remaining space on the screen, so that the footer is pushed to
+		// the bottom.
+		margin-bottom: auto;
+	}
 }
 
 // The following "placeholder selectors" define many of the layout-related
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-news-2021/sass/base/_layout.scss
```diff
@@ -1,14 +1,26 @@
...
 	display: flex;
 	flex-direction: column;
 	margin-top: -(var(--wp-admin--admin-bar--height, 0));
+}
+
+// Make sure the page grows to fill the browser window, and that the footer is stuck to the bottom.
+html,
+body,
+.wp-site-blocks {
+	height: 100%;
```
ryelle: This messes with the sticky header unfortunately - I had to fix a similar issue on translate (part of https://github.com/WordPress/wporg-mu-plugins/pull/96). You can test it by trying this out on a long page, once you scroll past the viewport height, the header unsticks.

iandunn: 👍🏻 , switching back to draft while I work on that

iandunn: I think that's solved in 5104cebbb7654842bc1c819fb90aea47355bf697

beafialho:
LGTM :+1:

ryelle (approved):
Tested on long & short pages, header still works and there's no space under the footer. 👍🏻

PR #249, opened 2022-01-26
Adjust vertical position of query title mask

mikachan:
This adjusts the vertical position of the query title mask by 1px, which should remove the grey line which is appearing sometimes at certain resolutions (e.g. 1440px).

| Before | After |
| ------ | ------ |
| image    |  image      |

Fixes #248.

PR #207, opened 2022-01-18
Try/refactor

shaunandrews:
Ok, so. I ended up deleting a bunch of CSS, reorganizing, and rewriting/refactoring just about... well... everything.

I know this is a big PR, and that it would be better if I broke up up into dozens (or more) smaller PRs — but I'm not sure its worth the extra effort. What I hope this PR does is create a solid foundation for us all to continue to improve. I'm sure I've done many things the "wrong way," or that there might be regressions from `trunk`, but overall I feel these changes are a net positive to the state of the theme.

PR #250, opened 2022-01-27
Modularize components/_local-header.scss

### LINKED ISSUE #52: Modularize `components/_local-header.scss`
It's too large to easily grok, and could probably be broken into several smaller logical pieces.

This was discussed in https://github.com/WordPress/wporg-news-2021/pull/41#discussion_r733958600, but wasn't completed before the merge.

- [ ] modularize
- [x] remove the `&-{partial class name}` pattern, to improve readability

mikachan:
This is a small PR that breaks down the CSS in `components/_local-header.scss` into multiple, smaller .scss files.

I went with creating a new directory called `local-header`, and then breaking down the contents into `local-header`, `breadcrumbs` and `categories`.

Closes #52.

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/components/local-header/_breadcrumbs.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/components/local-header/_breadcrumbs.scss
@@ -1,0 +1,33 @@
+body .wp-site-blocks .local-header {
+    // Breadcrumbs show the lineage of pages based on the current
+	// page. It contains a link for each parent.
+	.local-header__breadcrumb {
+		display: flex;
+		align-items: center;
+		min-height: var(--bar-min-height);
+		color: var(--bar-text-color);
+
+		// The current item is highlighted.
+		&-current {
+			font-family: var(--wp--preset--font-family--inter);
+			font-size: 1rem;
+			font-weight: 700;
+
+			// This ensures that the square separator
+			// is reliably centered.
+			display: flex;
+			align-items: center;
+			margin-top: 0;
+
+			&::before {
+				content: "";
+				height: 2px;
+				width: 2px;
+				background: var(--bar-text-color);
+				display: inline-block;
+				margin: 0 0.5rem;
+				border-radius: 2px;
+			}
+		}
+	}
+}
--- a/source/wp-content/themes/wporg-news-2021/sass/components/local-header/_breadcrumbs.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/components/local-header/_breadcrumbs.scss
@@ -1,31 +1,33 @@
-// Breadcrumbs show the lineage of pages based on the current
-// page. It contains a link for each parent.
-.local-header__breadcrumb {
-    display: flex;
-    align-items: center;
-    min-height: var(--bar-min-height);
-    color: var(--bar-text-color);
-
-    // The current item is highlighted.
-    &-current {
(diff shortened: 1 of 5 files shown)


source/wp-content/themes/wporg-news-2021/sass/components/local-header/_breadcrumbs.scss
```diff
@@ -0,0 +1,33 @@
+body .wp-site-blocks .local-header {
+    // Breadcrumbs show the lineage of pages based on the current
+    // page. It contains a link for each parent.
+    .local-header__breadcrumb {
```
ryelle: Can you simplify these selectors (in each file), or do they need to be this nested for specificity? `.local-header__breadcrumb` by itself seems like it would target the correct thing.

mikachan: Ah yeah, good spot. Looks like they can all be simplified. I've removed `body .wp-site-blocks .local-header` from each, apart from `local-header.scss` which I've changed to `.local-header`.

mikachan: Ah wait, I've just spotted an issue on the archive page with some conflicting margins in the breadcrumbs. Fixing..

mikachan: Looks like it was just the breadcrumbs that needed more specificity, so I've added `.local-header` there as well.

source/wp-content/themes/wporg-news-2021/sass/components/local-header/_breadcrumbs.scss
```diff
@@ -0,0 +1,33 @@
...
+        display: flex;
+        align-items: center;
+        min-height: var(--bar-min-height);
+        color: var(--bar-text-color);
+
+        // The current item is highlighted.
+        &-current {
```
iandunn:
```suggestion
        .local-header__breadcrumb-current {
```

#207 changed most of the partial selectors, but it looks like this one was missed.

iandunn (commented):
Thanks for making this easier to work with 🙂

mikachan:
Thanks @iandunn, I've committed that selector change as well.

source/wp-content/themes/wporg-news-2021/sass/components/local-header/_local-header.scss
```diff
@@ -0,0 +1,50 @@
...
+	color: var(--bar-text-color);
+	min-height: var(--bar-min-height);
+	position: relative;
+	background: linear-gradient(to top, transparent var(--bar-min-height), var(--bar-background-color) var(--bar-min-height) 100%);
+
+	&::after {
+		content: "";
+		min-height: var(--bar-min-height);
+		position: absolute;
+		bottom: 0;
```
ryelle:
```suggestion
		bottom: 1px;
```

Add the same fix here from #249, since I see the same 1px line in the mobile dropdown.

Screen Shot 2022-01-28 at 11 37 34 AM

ryelle (approved):
Looks good! I just made one suggestion.

Can you also run the linter before merging?

```
```

mikachan:
Thanks @ryelle! I've committed the same fix from #249, and I ran the CSS linter.

Issue #274, opened 2022-02-01
Posts: Media default width and caption alignment


beafialho:
I believe the current default width for posted images and video is content width. The [original intention](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=4659%3A11397) was to make them full-width by default, is that possible?

**Desktop example:**
<img width="749" alt="Captura de ecrã 2022-02-01, às 12 02 54" src="https://user-images.githubusercontent.com/65220155/151965200-dc98b4f0-dc6d-40f1-ba42-3d53b25a1b56.png">

**Mobile example:**
<img width="531" alt="Captura de ecrã 2022-02-01, às 12 02 44" src="https://user-images.githubusercontent.com/65220155/151965214-7e2c2dad-dc79-45a1-8859-18b3d5f874b8.png">

Currently, the image captions have extra side padding which should be adjusted to have it look as below.

**Currently:**
<img width="521" alt="Captura de ecrã 2022-02-01, às 11 55 20" src="https://user-images.githubusercontent.com/65220155/151964268-efb0076f-d139-4091-a985-5cc8cd684f66.png">

**Intended:**
<img width="522" alt="Captura de ecrã 2022-02-01, às 11 55 07" src="https://user-images.githubusercontent.com/65220155/151964279-c05b5f54-580e-4417-be37-5beaac3a9358.png">

If the full-width default issue isn't doable or too complicated to figure out, this is something we can come back to post-launch.

coreymckrill:
> I believe the current default width for posted images and video is content width. The original intention was to make them full-width by default, is that possible?

🤔 Since image and video blocks have a block layout setting for default|wide|full-width, I'm not sure we'd want to modify the styling so that `default` has the same width as `wide` or `full-width`. Because then if we actually wanted a specific block to be the normal content width, it wouldn't be possible. So my initial thought here is that this is a content issue rather than a style issue, which would require someone going through each post and changing all the image and video blocks to the `full-width` layout setting.

> Currently, the image captions have extra side padding which should be adjusted...

It looks like the selector that is causing the extra padding on the right side is: `[class*="wp-block-"].alignfull figcaption`

...Which is applying to all `figcaption`s in the content because the content container has an `alignfull` class on it. This might be fixed by adding specificity: `[class*="wp-block-"].alignfull > figcaption`

beafialho:
[...]

PR #275, opened 2022-02-01
Captions: Adjust side margins, remove boilerplate from Blockbase

coreymckrill:
* Removes side margins from figcaption elements, since the figure
  container will already have some whitespace around it.
* Except when the figure container is set to full-width, in which case
  some margin is applied to the right side.
* Removes custom styles for the video block that came over from
  Blockbase, but do not appear to be intended for the News site design.

Relates to #274

| Before | After |
| --- | --- |
| before | after |

| Full width |
| --- |
| full-width |

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/base/_elements.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/base/_elements.scss
@@ -29,12 +29,10 @@ a {
 	figcaption {
 		font-size: var(--wp--custom--gallery--caption--font-size);
 		text-align: right;
-		padding-left: var(--wp--style--block-gap);
-		padding-right: var(--wp--style--block-gap);
 	}
 
 	&.alignfull {
-		figcaption {
+		> figcaption {
 			padding-right: calc(var(--wp--custom--alignment--edge-spacing) + var(--wp--style--block-gap));
 		}
 	}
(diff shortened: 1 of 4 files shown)


coreymckrill:
@beafialho Is this what you had in mind in #274 for image caption margins?

beafialho:
Thank you @coreymckrill, the **After** screenshot above looks accurate. Let's add this caption rule to Image Compare blocks:

Captura de ecrã 2022-02-02, às 15 15 55

About full-width, is it possible to get it closer to the edge of the screen?

Captura de ecrã 2022-02-02, às 15 12 56

coreymckrill:
@beafialho 

> is it possible to get it closer to the edge of the screen?

Yep! The above commit makes it look like your mockup above.

> Let's add this caption rule to Image Compare blocks

Is this just referring to the caption being centered for this type of block? If so, the changes in this PR already make that the case.

beafialho:
> Is this just referring to the caption being centered for this type of block? If so, the changes in this PR already make that the case.

Sorry @coreymckrill I should have been clearer. The point is to have the same right-aligned captions for **image compare blocks** and **video blocks** as we have for images.

coreymckrill:
Ah, ok! No worries, I'll tweak the styles.

PR #277, opened 2022-02-02
Sort category menu by column instead of row

### LINKED ISSUE #146: Local Nav Bar: Rearrange Categories

tellyworth:
Fresh attempt to fix #146, replaces #243.

I can't figure out why the `flex-direction` in `break-small` seems to override the `column` sort. This breaks the vertical menu at small sizes, but has the correct sorting on wider screens.

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/block-template-parts/local-navigation.html
+++ b/source/wp-content/themes/wporg-news-2021/block-template-parts/local-navigation.html
@@ -26,7 +26,7 @@
 			</svg>
 		</label>
 		<!-- /wp:html -->
-		<!-- wp:categories /-->
+		<!-- wp:categories {"showOnlyTopLevel":true} /-->
 	</div>
 	<!-- /wp:group -->
 </div>
--- a/source/wp-content/themes/wporg-news-2021/sass/components/local-header/_categories.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/components/local-header/_categories.scss
@@ -54,7 +54,7 @@
 			}
 
 			~ .wp-block-categories {
-				display: flex;
+				display: block;
 			}
 		}
 	}
@@ -95,8 +95,6 @@
 		list-style: none;
 		padding: 0 0 30px 0;
 		display: none;
-		flex-direction: column;
-		flex-wrap: wrap;
 		max-height: none;
 
 		// Override the default margin applied to the block.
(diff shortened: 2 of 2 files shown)


dd32:
@tellyworth I think this can be achieved with CSS columns instead.

Replace 
```
.local-header__categories .local-header__categories-toggle:checked ~ .wp-block-categories {
    display: flex;
}
```

with:
```
.local-header__categories .local-header__categories-toggle:checked ~ .wp-block-categories {
    display: block;
    column-count: 5;
    white-space: nowrap;
}
```
(edit: Those might be the wrong selectors.. I didn't actually test it much other than in-browser editor)

It'll need some breakpoints defined to adjust it from 5 columns to 4.. 3.. 2 columns.
The mobile view can probably just be defined as a singular (or two) columns instead of flexing there too.

It also works better by dropping child categories:
`<!-- wp:categories {"showOnlyTopLevel":true} /-->`

tellyworth:
Fixed it - I had the breakpoints backwards.

Screenshots with the PR applied on my local test:
Screen Shot 2022-02-02 at 6 17 45 pm
Screen Shot 2022-02-02 at 6 17 56 pm

ryelle:
I had some trouble with the columns expanding offscreen when I tried this on my sandbox, so I tried @dd32's suggestion, and it works well. The only maybe-awkward thing is that different browsers use different algorithms to lay out the columns, it looks like Safari tries to keep the columns even, while Firefox fills up each column before moving to the next (so I had an orphan item in FF but not Safari). I think this is fine, only raising it in case someone notices it looks different 🙂 

| size | News |
|---|---|
| 480px, 1 column | 480-1-col |
| 650px, 2 columns | 650-2-col |
| 800px, 3 columns | 800-3-col |
| 960px, 4 columns | 960-4-col |
| 1440px, 5 columns | 1440-5-col |

(the header is a little broken, that's tracked in https://github.com/WordPress/wporg-mu-plugins/issues/142)

dd32:
> I had some trouble with the columns expanding offscreen when I tried this on my sandbox

That's what sent me down the CSS columns route :) 

> The only maybe-awkward thing is that different browsers use different algorithms to lay out the columns, it looks like Safari tries to keep the columns even, while Firefox fills up each column before moving to the next (so I had an orphan item in FF but not Safari).

I too ran into something similar, I had something akin to `5 4 5 4 4` items in each column in Chrome.. I couldn't see why it chose that, and the CSS rules that were supposed to control it didn't seem to do anything.

I think we're still going to need to do something to exclude rarely-used/only-old categories, which will reduce the item count which may make the UI look better.

dd32:
Looks good to me, although the non-even columns and orphaned elements are kind of annoying..

Unfortunately Firefox is the only browser that doesn't support the `widows` CSS property too: https://developer.mozilla.org/en-US/docs/Web/CSS/widows (although to be fair, Chrome/Safari don't appear to respect it for Column purposes)

ryelle:
> I think we're still going to need to do something to exclude rarely-used/only-old categories, which will reduce the item count which may make the UI look better.

Yeah, but that can be handled on the site itself. I'm going to merge this one now, and open another PR for the All Posts link.

PR #283, opened 2022-02-04
Remove unused custom css var definitions from theme.json and stylesheets

coreymckrill:
There were a few properties in the `custom` section of the theme.json file that either weren't being used or were pointing to other css variables that didn't exist. These were mostly color variable names that came over from Blockbase. This is just an effort to reduce clutter.

There should not be any visual changes on the site from these deletions.


ryelle (approved):
Nice cleanup! I don't see any changes on the site 👍🏻

PR #284, opened 2022-02-04
Local Nav: Add "All Posts" link to category listing

### LINKED ISSUE #71: Show "all posts" link in local nav category drop down
> The blue nav bar should only show the categories, "All Posts" being one of them:
> https://github.com/WordPress/wporg-news-2021/issues/55#issuecomment-953650339

I don't think the category block will do that by default, but there might be a reasonable workaround. If not, it's probably fine to leave it out.

I'm not sure what URL the link should go to, though, since `/page/1` would just redirect to `/` and show `home.html`; and `/page/2` would miss the posts from page 1.

Related #70

ryelle:
Fixes #71 — this filters the category list & adds a link to the "All Posts" page. Right now, "All Posts" is actually `/page/2`, the second page of posts. The first page of posts is the homepage (like #71 mentions). 

Screen Shot 2022-02-04 at 12 17 44 PM

If we need this page to be a list of all posts, including the first page, we could try creating a new page with a Query Loop block, but I don't want to go that route unless it's actually an issue 🙂

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/functions.php
+++ b/source/wp-content/themes/wporg-news-2021/functions.php
@@ -23,7 +23,8 @@ add_filter( 'render_block_data', __NAMESPACE__ . '\custom_query_block_attributes
 add_filter( 'template_redirect', __NAMESPACE__ . '\jetpack_likes_workaround' );
 add_filter( 'the_title', __NAMESPACE__ . '\update_the_title', 10, 2 );
 add_action( 'ssp_album_art_cover', __NAMESPACE__ . '\custom_default_album_art_cover', 10, 2 );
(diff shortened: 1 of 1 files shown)


beafialho:
> If we need this page to be a list of all posts, including the first page, we could try creating a new page with a Query Loop block, but I don't want to go that route unless it's actually an issue 

Thanks @ryelle! If that's possible, then the "All Posts" page should have all posts including the first page, as that's what a user would expect.

coreymckrill (approved):

ryelle:
> If that's possible, then the "All Posts" page should have all posts including the first page, as that's what a user would expect.

Okay - I think I have something working locally, but I'll merge this PR and create a new one for the new page, so we can update both the See All Posts & All Posts links.

Issue #251, opened 2022-01-27
Launch tasks


iandunn:
### Before launch

- [x] draft announcement post. ask for feedback. where to publish? xpost to relevant p2s.
- [x] tweet about it coming next week
- [x] notify `#teamreps` `#design` `#meta` channel that launch will happen soon
- [x] specifically notify Make Marketing as well as Community and Core (in their meetings)
- [x] maybe do some beta testing w/ contributors?
- [x] Green light from Neso (clear to launch on tech issues)
- [x] Green light from Design (clear to launch on design completion)
- [x] Green light from Navigators (announcements, timing)

### Launch day

- [x] one last final 👍 from @beafialho 
- [x] sync svn from git, deploy
- [x] Activate the new theme on the production site
- [x] Update Reading settings to have static home and posts pages
- [x] Activate gutenberg on production site? done but [deactivated](https://github.com/WordPress/wporg-mu-plugins/issues/92#issuecomment-1019688072) and re-activated.
- [x] update [news ID in `save_dependent_global_styles`](https://github.com/WordPress/wporg-mu-plugins/blob/11225d2d5f895414f40904c169a66f882cb8113a/mu-plugins/blocks/global-header-footer/blocks.php#L907) when activating news theme. use `WPORG_NEWS_BLOGID`
- [x] Remove the preview toggle code in `mu-plugins/main-network/site-news.php`
- [x] publish announcement post
- [x] marketing markets post

### Clean up

- [x] make sure no code is referring to w.org/news-test anymore
- [x] delete w.org/news-test ?
    - [x] if so, remove temporary News link in `get_menu_url_for_current_page()`

please add more items if you think of anything else

tellyworth:
I'd add these as prerequisites for launch:

* Green light from Neso (clear to launch on tech issues)
* Green light from Design (clear to launch on design completion)
* Green light from Navigators/Marketing? (clear to launch on externals, timing, announcements etc)

kellychoffman:
Announcement post draft: https://docs.google.com/document/d/1xxyOa8rc7mctXU_RQpZ-SRW6rjIRYFV0Y6qgP_lhJLI/edit

PR #287, opened 2022-02-04
Add an "All Posts" template to list full archive

ryelle:
See #71, #284. Switch homepage and front page to use static pages, so that we can have a separate "All Posts" page. This also simplifies some of the logic around template loading.

New testing instructions in https://github.com/WordPress/wporg-news-2021/pull/287#issuecomment-1031885548

<details>
<summary>Old description</summary>
This adds a new template which is set on a page called "All Posts". When used, it shows a list of all the posts on the site in a simple list, 10 posts per page. It uses the same design & class names as the existing paged archive, but this one also shows the first page of posts too.

To test:

- Create a new page called "All Posts", the slug should be `all-posts` for the links to work
- Set the page template to "All Posts", don't worry about content. Publish it.
- View the page, it should show you a list of the latest posts.
- Use the pagination, it should correctly nav through older posts
- The "All Posts" item in the categories nav should stay highlighted

Note: I didn't remove any of this style logic from the `/page/n` pages, since power users might still visit those, they will look the same.
</details>

<details>
<summary>Full page screenshot</summary>

Screen Shot 2022-02-04 at 15 46 47

</details>


coreymckrill (commented):
🤔 This works as described... but there are some quirks. Using the pagination links gives you URLs that look like this: `http://localhost:8888/all-posts/?query-15-page=2`. Not sure what the `query-15` part is about. Trying to do `http://localhost:8888/all-posts/page/2` doesn't seem to work, it shows the same posts no matter what.

A more complicated alternative might be to set up static pages for the homepage and posts pages. I _think_ that would do the same as this, but also allow for `http://localhost:8888/all-posts/page/2` URLs. But there would probably need to be some styling tweaks. I briefly tried this and the layout was broken 😕

ryelle:
> Not sure what the query-15 part is about.

That's how pagination works in the Query Loop block - `query-15` comes from [the queryId here](https://github.com/WordPress/wporg-news-2021/blob/2f58d14ba87743e10cd96cdf9e6431ca988acd17/source/wp-content/themes/wporg-news-2021/block-templates/all-posts.html#L5). Because there could be more than one query loop on the page, the URL needs to identify which query to paginate.

> A more complicated alternative might be to set up static pages for the homepage and posts pages.

I think we probably should try that then, since this also has the awkward side-effect that you can still go to `http://localhost:8888/page/2/` and paginate from there.

ryelle:
Okay, I've changed the logic here. Instead of a new page + page template, this now uses a static page for the homepage, and the All Posts page for the posts page.

This also simplifies the paged template loading, so I was able to remove `override_front_page_template` & `offset_paginated_index_posts`, and our custom `is_posts_index` function (we can use `is_home` for that now).

To test:

- Add pages for "Home" & "All Posts"
- Set Homepage to "Home", and posts page to "All Posts" in Options > Reading
- View the site, it should work as expected.
- Clicking "See all posts" on homepage will direct you to `/all-posts/`
- The "All Posts" item in the category nav should be highlighted

The only issue with merging this is that the preview mode on `w.org/news/` won't work anymore, since it requires a database change to set the front page. It'll also add a step to launch #251.

source/wp-content/themes/wporg-news-2021/functions.php
```diff
@@ -141,86 +139,6 @@ function modify_archive_title_prefix( $prefix ) {
...
 
-/**
- * Load the `index.html` template for `w.org/news/page/{n}` requests, rather than `front-page.html`.
- *
- * The design calls for a mix of `front-page.html` and `home.html`/`index.html` functionality. "home" is Core's
- * legacy terminology for the posts index from the pre-CMS days. The design calls it the "all posts" screen.
- *
- * Setting it up in separate files requires this hack, but feels more straight-forward overall. This lets us have
- * static content and the latest posts on the front page, while also preserving clean URLs. Using
- * `show_on_front = page` would change the URLs to `w.org/news/posts/page/2` rather than `w.org/news/page/2`.
```
ryelle, line 153: It's true that now the paged URLs are `/news/all-posts/page/2/`, instead of just `/news/page/2/`. Is this still considered a problem?

coreymckrill: I think it's fine. Having a slug URL structure instead of a query string is the part I was most concerned about.

coreymckrill (approved):
👍 Looks good to me!

> The only issue with merging this is that the preview mode on w.org/news/ won't work anymore, since it requires a database change to set the front page. It'll also add a step to launch https://github.com/WordPress/wporg-news-2021/issues/251.

I'm not sure how the preview mode is implemented, but could it be extended to filter the values of the relevant `wp_options` entries, so that in preview we have static front/blog pages, and in non-preview we don't?

Either way, doesn't seem like too big of a deal, and I do think this is a better approach.

dd32:
This change looks good to me from a code-review perspective 👍 

> I'm not sure how the preview mode is implemented, but could it be extended to filter the values of the relevant wp_options entries, so that in preview we have static front/blog pages, and in non-preview we don't?

Yep, can be done. See `mu-plugins/main-network/site-news.php` around line 155, add something akin the following after the template/stylesheet filter.
```
add_filter( 'pre_option_show_on_front', function() { return 'page'; } );
add_filter( 'pre_option_page_on_front', function() { return 123; } );
add_filter( 'pre_option_page_for_posts', function() { return 456; } );
```
I'll do that the day after this is merged if someone else doesn't :)

beafialho:
Thanks @ryelle, when I test this PR and click "See All Posts" in the homepage or in the local nav dropdown, this is what I see. Is that expected or am I doing something wrong?

https://user-images.githubusercontent.com/65220155/152962899-e7a14eeb-7402-4913-96e9-0470056c2eeb.mp4

dd32:
@beafialho You'll need to define a Static front page & Page for posts in wp-admin first.

See the list in this comment: https://github.com/WordPress/wporg-news-2021/pull/287#issuecomment-1031885548 + https://wordpress.org/support/article/creating-a-static-front-page/

beafialho:
@dd32 I managed to do that, but I can't seem to see the pages properly. 

Anyway, since this likely doesn't have any visual concerns besides the ones added [here](https://github.com/WordPress/wporg-news-2021/issues/296), I trust your review that this links as intended 👍

ryelle:
> I managed to do that, but I can't seem to see the pages properly.

The code expects the page you created to be called All Posts, with the slug `all-posts` (like http://localhost:8888/all-posts/). If you named it something else, those links won't work. In any case, I'll merge this & update the `news-test` site so you can double-check on something live-ish 🙂

beafialho:
> The code expects the page you created to be called All Posts, with the slug all-posts (like http://localhost:8888/all-posts/). If you named it something else, those links won't work. In any case, I'll merge this & update the news-test site so you can double-check on something live-ish 

Awesome, thanks!

PR #278, opened 2022-02-02
Add padding and margin to the pagination block

tellyworth:
Fixes #268 spacing.

PR #288, opened 2022-02-04
Fix spacing on non-category archives

ryelle:
This fixes two issues on non-category archives (dates, tags, etc)

- On screens smaller than 1280px, there was no space around the content
- In the local nav breadcrumbs, the archive title had no space between the label & title

| Type | Before | After |
|------|--------|-------|
| Category (no change) | before-cat | after-cat |
| Date | before-date | after-date |
| Tag | before-tag | after-tag |

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/post/_content.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/post/_content.scss
@@ -26,6 +26,7 @@
 body.archive .site-content-container {
 	padding-left: var(--wp--custom--alignment--edge-spacing);
 	padding-right: var(--wp--custom--alignment--edge-spacing);
+	padding-bottom: var(--wp--custom--alignment--edge-spacing);
 }
 
 body.news-posts-index .site-content-container,
--- a/source/wp-content/themes/wporg-news-2021/sass/post/_content.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/post/_content.scss
@@ -23,6 +23,11 @@
 	}
 }
 
+body.archive .site-content-container {
+	padding-left: var(--wp--custom--alignment--edge-spacing);
+	padding-right: var(--wp--custom--alignment--edge-spacing);
+}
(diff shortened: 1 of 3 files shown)


coreymckrill (commented):
👍 Works, except it's not fixing the issue for All Posts pages, e.g. http://localhost:8888/page/2

...because `body` doesn't have an `archive` class on it in that case.

But maybe that wouldn't be necessary anyway after #287 ?

ryelle:
> But maybe that wouldn't be necessary anyway after https://github.com/WordPress/wporg-news-2021/pull/287 ?

Yeah, #287 fixes it for both the `paged` template and the `all-posts` custom page template, since neither of them are technically `archives`. WordPress 🤷🏻

source/wp-content/themes/wporg-news-2021/sass/post/_content.scss
```diff
@@ -23,6 +23,11 @@
 	}
 }
 
+body.archive .site-content-container {
+	padding-left: var(--wp--custom--alignment--edge-spacing);
+	padding-right: var(--wp--custom--alignment--edge-spacing);
+}
```
coreymckrill:
```suggestion
	padding-right: var(--wp--custom--alignment--edge-spacing);
	padding-bottom: var(--wp--custom--alignment--edge-spacing);
}
```

I think this would address the spacing issue when there's no pagination mentioned [here](https://github.com/WordPress/wporg-news-2021/pull/278#issuecomment-1028066128).

coreymckrill: (I could also add this in a separate PR once this one is merged.)

ryelle: I was just testing it 🙂  It does fix the issue, and adds more space under the pagination which better matches the design too

| Before | After | Design |
|--------|------|-------|
| before | after | design |

PR #290, opened 2022-02-04
Add "Month in WordPress Title" block

### LINKED ISSUE #269: Month in WordPress: what's the best logic?
In the Month in WordPress category, when I click the December 2021 post, my expectation is to read the post about December. Instead, it takes me to the post about November. I get that it works this way because it links to _when_ it was posted, but that can be confusing (at least I find it misleading).

Does this make sense to everyone as well and if so, is there a way to change the way these links work?

https://user-images.githubusercontent.com/65220155/151838824-12c41c16-c539-49de-89b0-b3342b968a43.mp4

ryelle:
Adds "Month in WordPress Title" block, which renders a heading (optionally with a link to the post). It outputs the short month name with an `aria-label` of "full month", and includes the year in the heading, so users browsing by heading have more context.

I also tightened up the spacing & font sizes to better match the design in figma.

**Screenshots**

1350px wide (this is the largest the text size gets)

800px wide

1350px wide (this is the smallest the text size gets)

**To test**

1. View the Month in WordPress category
2. It should match the design
3. Try navigating with a screen reader, the heading/links should be understandable, ex "March 2021"


beafialho:
This looks great @ryelle! Only one comment, can you decrease the stroke size a bit on mobile so it looks the same width as the grid of posts?

| Currently | Figma Mockup |
| ----- | ----- |
|    Captura de ecrã 2022-02-09, às 09 53 53    |    Captura de ecrã 2022-02-09, às 09 54 21   |

source/wp-content/themes/wporg-news-2021/blocks/month-in-wp-title/index.php
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	$post_id  = $block->context['postId'];
+	$tag_name = 'h2';
+
+	if ( isset( $attributes['level'] ) ) {
+		$tag_name = 0 === $attributes['level'] ? 'p' : 'h' . $attributes['level'];
+	}
+
+	$date = date_create_from_format( 'Y-m-d H:i:s', get_the_date( 'Y-m-00 00:00:00', $post_id ) );
```
coreymckrill, line 32: `'Y-m-00 00:00:00'`

Is this what makes it grab the _previous_  month instead of the month that the post was actually published in? If so, I think it would be worth documenting that in a code comment, as it doesn't stand out that that's what's happening here.

source/wp-content/themes/wporg-news-2021/blocks/month-in-wp-title/index.php
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Register block type in JS, for the editor.
+ */
+function register_block_type_js() {
+	$block = wp_json_file_decode( __DIR__ . '/block.json' );
+	ob_start();
+	?>
+	( function( wp ) {
+		wp.blocks.registerBlockType(
+			'<?php echo esc_html( $block->name ); ?>',
```
coreymckrill: 🤔 Should these be `esc_js` instead of `esc_html`?

coreymckrill (approved):
👍 Working well, seems like an elegant solution. Two minor comments, but not blockers.

source/wp-content/themes/wporg-news-2021/blocks/month-in-wp-title/index.php
ryelle: Yep, that's what's going on. Updated with a comment in both files this appears https://github.com/WordPress/wporg-news-2021/pull/290/commits/f8d5054833274d0d5e778f9e41a89f852d6542d2

source/wp-content/themes/wporg-news-2021/blocks/month-in-wp-title/index.php
ryelle: I always forget that the other `esc_`s exist 😅  — updated in https://github.com/WordPress/wporg-news-2021/pull/290/commits/e72015ec8f031c78c87848bee3d3ed7ebc5d9206

ryelle:
@beafialho I've updated the stroke size a bit, so it looks like this now:

coreymckrill:
Is this ready to go? Would love to get this merged so that I can move the block code in #246 into the new blocks folder created here.

ryelle:
Yeah, I think we can merge it. If @beafialho has any more feedback about the stroke we can handle that in another PR.

beafialho:
Sorry for the delay, this looks good!

PR #303, opened 2022-02-11
Allow the `last-in-year` class to be added to the first post.

dd32:
Fixes #301

Testing instructions:

 - Have the first post in the loop be the last post of a year.
 - visit /category/month-in-wordpress/
 - See the first post followed by a separator, followed by last years posts.

Screen Shot 2022-02-11 at 2 16 36 pm

PR #306, opened 2022-02-11
Remove the `first-in-year` class name

ryelle:
The `first-in-year` class is not used in the theme, since only the `last-in-year` is needed for the Month in WordPress break. Since it's not used, we can remove the logic for adding it to the posts. There should be no visual change.

To test:

- View the `month-in-wordpress` category
- Check that `first-in-year` is no longer added to the January posts
- Check that `last-in-year` is still added to December posts
- There should be no visual changes


coreymckrill (approved):

dd32 (approved):
Agreed, LGTM.

I left the code in place in #303 as it worked, and I figured it might be needed for another use-case of the block.

PR #312, opened 2022-02-14
Community Category: Set a 3-line maximum on post titles.

### LINKED ISSUE #300: Community category: Long post titles overflow out the top of the post box for some viewport widths
On the Community category archive, at some screen widths the post titles start to overflow out the top of their post boxes. Here it is at 960px:

text-overflow

I tooled around with the CSS but couldn't find a good way to truncate the text in a way that looked decent. Maybe scaling the font down a bit more would help, but it seems like there will always be an edge case with a long title that will need to be handled somehow.

cc @beafialho

ryelle:
Fix #300 — this sets a max-height of 3 lines for the post title, and hides the rest. It also adjusts the font size to be 26px at 960px wide, and scales up to 32px from there.

~It doesn't add any sort of ellipsis to indicate the title is cut off - I tried adding one as `after` content but it sits at the end of the text (cut off), and tried absolute positioning, but then it always sits at the bottom right regardless of where the end of the line is horizontally. I thought this solution was enough improvement to make a PR, though.~ With `line-clamp`, now it does add the ellipsis 🙂 

<details>
<summary>More screenshots at different sizes</summary>

480

650

790

1280

1440

</details>

To test:

- View the community category across sizes.
- The post titles shouldn't overflow out of the square, and they shouldn't visible clip (cut off) either.

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/page/_categories.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/page/_categories.scss
@@ -173,8 +173,10 @@ body.category-community {
 	}
 
 	.wp-block-post-title {
-		font-size: var(--wp--preset--font-size--huge);
+		font-size: clamp(26px, 2vw, 32px);
 		line-height: 1.3;
+		max-height: 3.9em;
+		overflow: hidden;
 
 		a:hover,
 		a:focus {
--- a/source/wp-content/themes/wporg-news-2021/sass/page/_categories.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/page/_categories.scss
@@ -175,8 +175,13 @@ body.category-community {
 	.wp-block-post-title {
 		font-size: clamp(26px, 2vw, 32px);
 		line-height: 1.3;
-		max-height: 3.9em; // 3 lines: 3em * 1.3 line height.
-		overflow: hidden;
+
+		a {
+			display: -webkit-box;
+			-webkit-box-orient: vertical;
+			-webkit-line-clamp: 3;
+			overflow: hidden;
+		}
 
 		a:hover,
 		a:focus {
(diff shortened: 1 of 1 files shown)


beafialho:
Thank you @ryelle, this is looking great! The only size where it definitely should be squared (or at least relevant spacing above the titles) is at mobile size. It looks a bit tight the way it is:

Currently:
Captura de ecrã 2022-02-14, às 18 17 33

Intended:
Captura de ecrã 2022-02-14, às 18 17 02

Also, is there a way to make the images have a fixed height? As in:

Captura de ecrã 2022-02-14, às 18 17 58

Having them vary a lot in height makes it more confusing at this size.

coreymckrill (approved):
👍 Works great testing on local

ryelle:
Updated the aspect ratio code to make sure everything's square on small screens too, how's this @beafialho?

Screen Shot 2022-02-14 at 15 31 27

zackkrida:
We've recently started using the `line-clamp`[^1] css property in Openverse for truncating text. Here's some example code:

```css
.line-clamp-3 {
	overflow: hidden;
	display: -webkit-box;
	-webkit-box-orient: vertical;
	-webkit-line-clamp: 3;
}
```

Here it is as applied _to your GitHub comment!_

CleanShot 2022-02-14 at 18 15 28@2x


[^1]: https://caniuse.com/css-line-clamp

beafialho:
> Updated the aspect ratio code to make sure everything's square on small screens too, how's this @beafialho?

Thank you @ryelle, it looks good!

ryelle:
@zackkrida oh neat! I hadn't heard of that property. I switched to that so we get the ellipsis:

Screen Shot 2022-02-15 at 11 20 28 AM

PR #314, opened 2022-02-14
Refine 404 styles

### LINKED ISSUE #311: Implement 404 Page according to the design
[Figma link](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=4659%3A15520)

Captura de ecrã 2022-02-14, às 12 41 04

iandunn:
See #311

Doesn't try to fix these, since they'll be easier to focus on in separate PRs, after this is merged.

* the negative offset on the "oops" element. I ran into a bunch of weird overflow issues
* the `home-link` block empty output. seems like a recent regression, it's happening on trunk, and the cause isn't obvious. maybe related to recent gutenberg release.

Screen Shot 2022-02-15 at 08 55 20-fullpage

Screen Shot 2022-02-15 at 08 56 12-fullpage

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/block-templates/404.html
+++ b/source/wp-content/themes/wporg-news-2021/block-templates/404.html
@@ -10,13 +10,13 @@
 	<h1>This page doesn't exist.</h1>
 	<!-- /wp:heading -->
 
-	<!-- wp:html -->
-	<div>
+	<!-- wp:paragraph -->
+	<p>
 		Go to
-		<ul class="is-inline"><!-- wp:home-link {"label":"the homepage","tagName":"div"} /--></ul>,
+		<a href="https://wordpress.org/news/">the homepage</a>,
 		or try searching News posts from the field below.
-	</div>
-	<!-- /wp:html -->
+	</p>
+	<!-- /wp:paragraph -->
 
 	<!-- wp:search {"label":"","placeholder":"Search...","buttonText":"Search","buttonUseIcon":true,"buttonPosition":"button-inside"} /-->
 </main>
--- a/source/wp-content/themes/wporg-news-2021/block-templates/404.html
+++ b/source/wp-content/themes/wporg-news-2021/block-templates/404.html
@@ -1,24 +1,25 @@
 <!-- wp:template-part {"slug":"header","align":"full","className":"site-header-container"} /-->
 
-<!-- wp:group {"layout":{"inherit":true},"tagName":"main"} -->
+<!-- wp:group {"tagName":"main","className":"site-content-container"} -->
 <main class="wp-block-group site-content-container">
+	<p class="error404__oops" aria-hidden="true">
+		Oops!
+	</p>
 
-<!-- wp:heading {"level":1,"fontSize":"large"} -->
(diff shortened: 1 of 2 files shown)


beafialho:
Thanks @iandunn, I can't see it properly because my testing env is somehow stealing some space below the header, but besides the "Oops" element, the screenshots looks accurate 👍

source/wp-content/themes/wporg-news-2021/block-templates/404.html
```diff
@@ -1,24 +1,25 @@
...
-<div>
-	Go to
-	<ul class="is-inline"><!-- wp:home-link {"label":"the homepage","tagName":"div"} /--></ul>,
-	or try searching from the field below.
-</div>
-<!-- /wp:html -->
+	<!-- wp:html -->
+	<div>
+		Go to
+		<ul class="is-inline"><!-- wp:home-link {"label":"the homepage","tagName":"div"} /--></ul>,
```
ryelle: Could this be a plain `a` tag instead of `home-link`? It's already in an HTML block, we don't need to stick to editor-renderable blocks. That would avoid needing the `ul` wrapper which adds awkward semantics for screen readers.

ryelle:
> the home-link block empty output. seems like a recent regression, it's happening on trunk, and the cause isn't obvious. maybe related to recent gutenberg release.

I think you're behind on a Gutenberg update, or you've got it deactivated locally.

ryelle (commented):
Overall this looks good, just a few comments, mostly accessibility-adjacent.

Screen Shot 2022-02-15 at 1 06 55 PM

The "homepage" link here should be using `var(--wp--preset--color--blue-3);`, and the search input could use a focus state (I don't know if there's one designed, @beafialho).

iandunn:
> I think you're behind on a Gutenberg update, or you've got it deactivated locally.

Ah, you're right, thanks!

I had Git and SVN checkouts of it for various reasons, but accidentally updated it w/ wp-cli, which left the folder empty except for `node_modules`.

source/wp-content/themes/wporg-news-2021/block-templates/404.html
iandunn: 🤔  it would be nice to simplify it, but I can't think of a HTML-only URL that would be correct for local, staging, and prod.

`<a href="/news/">` would be a 404 on local, and the prod site on staging.

xref https://github.com/WordPress/gutenberg/issues/35926

We could write a custom block if you think the awkward semantics are bad enough to justify it.

iandunn: Well, maybe having the URLs wrong on local/staging is a smaller problem than the alternatives. That'd work for me.

iandunn: changed in https://github.com/WordPress/wporg-news-2021/pull/314/commits/7bd6a80b0eecf72ae3d9bf0e30184f9efa8cc216

iandunn:
I think the only remaining thing is possibly the focus state. I'm going to merge this to make it easier to start on the follow-up PRs, and the focus state can be done in one of those based on Beatriz's feedback.

PR #322, opened 2022-02-16
Use locally hosted fonts instead of Google Fonts

### LINKED ISSUE #234: Use locally-hosted version of Inter instead of Google Fonts
The Google Fonts version of Inter doesn't include italics: https://fonts.google.com/specimen/Inter

As a result, italic text will be faked by the browser by slanting the letters and that's not a good solution. Instead of sticking with this older version, we should switch to using the variable version from here, which we would bundle with the theme: https://github.com/rsms/inter/

ryelle:
Fixes #234. Requires https://github.com/WordPress/wporg-mu-plugins/pull/161, which has been merged. The new mu-plugin provides locally-hosted versions of EB Garamond & Inter, so we can remove the google fonts loading step here, and use the global-fonts stylesheet instead.

**To test**

- With the new mu-plugin active, load up News
- There should be no network request to `fonts.googleapis.com`
- The fonts should look the same as trunk
- Edit a post or page
- The fonts should load in the editor too
- You might still see `Inter@500,600` being loaded from google — that's from the Stream plugin.

⚠️  Note that the fonts won't load in the editor on localhost — this has to do with how [the styles are loaded into the editor](https://core.trac.wordpress.org/browser/trunk/src/wp-includes/block-editor.php#L528), using `wp_remote_get`. In `wp-env`, I get an error about connecting to `localhost:8888`. It works correctly on sandboxes.

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/functions.php
+++ b/source/wp-content/themes/wporg-news-2021/functions.php
@@ -3,6 +3,7 @@
 namespace WordPressdotorg\Theme\News_2021;
 
 use WP_Query;
+use function WordPressdotorg\MU_Plugins\Global_Fonts\get_font_stylesheet_url;
 
 defined( 'WPINC' ) || die();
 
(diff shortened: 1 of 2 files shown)


paaljoachim:
Font related:
Simplified Webfonts API
https://github.com/WordPress/gutenberg/pull/37140

coreymckrill (approved):
👍 Looks great on the sandbox!

PR #323, opened 2022-02-16
 404: Offset "oops" to match design

### LINKED ISSUE #311: Implement 404 Page according to the design

iandunn:
Fixes #311

Follow up to #314. 

* Moves "oops" to be partially off screen.
* removes an invalid `margin-top` that wasn't affecting anything
* standardizes the local header height var
* adds specificity to the search form styles, to prevent them bleeding into the global header search

The black footer logo is just something off in my environment (also on `trunk`). I don't think it'll affect staging/prod.

Screen Shot 2022-02-16 at 09 08 47-fullpage

Screen Shot 2022-02-16 at 09 08 57-fullpage

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/block-templates/404.html
+++ b/source/wp-content/themes/wporg-news-2021/block-templates/404.html
@@ -14,7 +14,7 @@
 	<p>
 		Go to
 		<a href="https://wordpress.org/news/">the homepage</a>,
-		or try searching News posts from the field below.
+		or try searching News posts using the field below.
 	</p>
 	<!-- /wp:paragraph -->
 
--- a/source/wp-content/themes/wporg-news-2021/sass/base/_layout.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/base/_layout.scss
@@ -4,7 +4,6 @@
 	min-height: 100vh;
 	display: flex;
 	flex-direction: column;
-	margin-top: -(var(--wp-admin--admin-bar--height, 0));
 
 	> .site-content-container {
 		// Make the content area grow to fill any remaining space on the screen, so that the footer is pushed to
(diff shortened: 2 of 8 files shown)


ryelle:
The footer logo isn't just your env, it's an issue from the logo change yesterday. See https://github.com/WordPress/wporg-mu-plugins/issues/162

ryelle (approved):

beafialho:
Thank you, on mobile it looks good 👍 

On desktop, it looks like there's too much space below the content, can we match that with the design? Also, the search field seems to be too high, let's make it `40px` high, as shown on Figma.

| Screenshot | Mockups |
| --- | --- |
| Screenshot |  Mockup |

iandunn:
I reduced the gap a bit in 74ca6fa (not deployed yet), but it's difficult to get it exact at every breakpoint, because of the dynamic font size.

I don't think most of the extra gap is coming from that, though. When the viewport is bigger than the page content, the footer sticks to the bottom, so the extra gap is expected to some degree.

iandunn:
a9fd18e fixes the search spacing. i'm going into a meeting, but will deploy it after that

iandunn:
Looks like @ryelle synced and deployed it, thanks!

@beafialho , you can see this on production now. LMK if you'd like any other changes.

beafialho:
Thank you @iandunn, there's still too much space between the content and the footer, let's try to make the content sit in the middle of the page so that there's similar space above and below:

Captura de ecrã 2022-02-18, às 11 02 32

iandunn:
Ah, ok, I didn't realize you wanted it vertically centered. I'll look into that.

iandunn:
I got that mostly working in 388a4f04ab48a02a737a0b290f83a3e9b98a45f9. It stops working once the viewport is `1050px` tall, because of the sticky footer issue mentioned above. I couldn't find a good way around that. It still looks good at those sizes, though, just different.

PR #328, opened 2022-02-17
Use post-author-name block so name is linkable, link to wporg profile

### LINKED ISSUE #257: Author name does not link
On the current site design, author names on posts link to their WordPress.org profiles. This can be really helpful for seeing some additional context about who the person is, their role in the community, and other things they've worked on.

coreymckrill:
Swaps out the `post-author` block (which allows for showing the author's avatar, but not for linking to their archive), for the `post-author-name` block (which allows for linking, but not showing the avatar). We're not using the authors' avatars anyway. Then filters the `author_link` hook to replace the archive URL with the author's wp.org profile page URL.

This also removes some unused/unnecessary styles related to the `post-author` block.

Fixes #257

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/block-template-parts/content-page.html
+++ b/source/wp-content/themes/wporg-news-2021/block-template-parts/content-page.html
@@ -5,7 +5,7 @@
 		<!-- wp:group {"className":"entry-meta"} -->
 		<div class="wp-block-group entry-meta">
 			<!-- wp:post-date /-->
-			By <!-- wp:post-author {"showAvatar":false} /-->
+			By <!-- wp:post-author-name {"isLink":true} /-->
 		</div>
 		<!-- /wp:group -->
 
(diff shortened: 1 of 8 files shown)


source/wp-content/themes/wporg-news-2021/sass/blocks/_post-author.scss
```diff
@@ -1,3 +0,0 @@
-.wp-block-post-author__name {
-	font-weight: var(--wp--custom--post-author--font-weight);
```
ryelle, line 2: This value can also be removed from `theme.json`, since it's not used anywhere else.

ryelle (approved):
Looks good, authors are showing up as expected and link to the profiles. 👍🏻

source/wp-content/themes/wporg-news-2021/sass/blocks/_post-author.scss
coreymckrill: Good catch!

coreymckrill: https://github.com/WordPress/wporg-news-2021/pull/328/commits/40ac25689a864d657212e0c9de56575de02ea701

Issue #95, opened 2021-11-30
Unexpected/invalid block errors in Site Editor mode


tellyworth:
The global header and footer blocks produce error output when viewing pages in Site Editor mode:
<img width="807" alt="Screen Shot 2021-11-30 at 5 35 01 pm" src="https://user-images.githubusercontent.com/7200686/143997804-c49fee0b-ec1e-439d-bc34-4f9456696a03.png">

I think this is probably due to the lack of `editor_script` values for those blocks (which are dynamic PHP blocks).

It's a bit of a pain to add a script easily given the complexity of the global header stuff, so I haven't tested that to confirm.

They don't actually need an editor script, just a placeholder will do (assuming that is indeed what's causing the errors).

Related: #7.

tellyworth:
@ryelle this one isn't needed till after SotW but I'd like to hear your ideas for how it ought to be fixed.

ryelle:
There are two issues here, as far as I can tell— the missing header/footer blocks, and a few block validation errors. I fixed the missing header & footer blocks with https://github.com/WordPress/wporg-mu-plugins/pull/26. The block validation errors seem to be due to the developer comments in the HTML of the template files.

Removing the comments would fix the issues, but they're useful to explain reasoning… so I'm not totally sure what the best action would be.

iandunn:
IMO Gutenberg should allow comments, for this reason. If needed, there could be a reserved `<!-- wp:comment ... -->` format, that gets ignored by the parser. We could open an issue upstream for that.

In the mean time, I don't think it's a priority, since we don't expect anyone to be editing the templates in practice -- see #14 . That could be desired in the future, though.

Issue #57, opened 2021-10-26
Gutenberg: Replace hardcoded URLs


iandunn:
There are several places we've hardcoded production URLs, because Gutenberg doesn't provide a way to create navigation links, category links, RSS feed URLs, etc.

Once there is a way, we should update them, so that they're accurate in local environments.

Related https://github.com/WordPress/gutenberg/issues/35926,  https://github.com/WordPress/gutenberg/issues/32939, https://github.com/WordPress/gutenberg/issues/31815#issuecomment-948117731

- [ ] hardcoded urls like in `html` blocks (e.g., `block-template-parts/home/`)
- [ ] `home-link` block in `404.html`
- [ ] Grep the codebase for others

PR #330, opened 2022-02-17
Link to the podcast category in single podcast.

StevenDufresne:
This PR links the `Podcast` text to the `/news/podcast` URL. This is not a great approach especially because `news` is hard coded and our local dev doesn't serve out of a subfolder.

I'm getting this PR early to discuss the best approach. Is there a better way to link to urls in HTML templates?

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/block-templates/single-podcast.html
+++ b/source/wp-content/themes/wporg-news-2021/block-templates/single-podcast.html
@@ -6,7 +6,7 @@
 	<div class="wp-block-group entry-meta">
 		<!-- wp:post-date /-->
 		By <!-- wp:post-author {"showAvatar":false} /-->
-		in Podcast
+		in <a href="/podcast">Podcast</a>
 	</div>
 	<!-- /wp:group -->
 
--- a/source/wp-content/themes/wporg-news-2021/block-templates/single-podcast.html
+++ b/source/wp-content/themes/wporg-news-2021/block-templates/single-podcast.html
@@ -6,7 +6,7 @@
 	<div class="wp-block-group entry-meta">
 		<!-- wp:post-date /-->
 		By <!-- wp:post-author {"showAvatar":false} /-->
-		in <a href="/podcast">Podcast</a>
+		in <a href="/news/podcast">Podcast</a>
 	</div>
 	<!-- /wp:group -->
 
--- a/source/wp-content/themes/wporg-news-2021/sass/post/_podcast.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/post/_podcast.scss
@@ -39,6 +39,9 @@ body.post-type-archive-podcast {
 			}
 		}
 
+		a {
+			color: var(--wp--preset--color--white);
+		}
 	}
 
 	.left-column {


iandunn:
Unfortunately that's a shortcoming of FSE. #57 has some background info

We've used relative URLs in a few places, but IIRC that only works on the homepage:

https://github.com/WordPress/wporg-news-2021/blob/8f7886fadff92a6904d4d1cef1158383002b0d27/source/wp-content/themes/wporg-news-2021/block-template-parts/front-page/latest-posts.html#L25

I think in this case hardcoding the production URL is the least-bad option, unless someone else knows of a better way?

ryelle:
Podcasts isn't actually a category, or else we could use `<!-- wp:post-terms {"term":"category"} /-->`. I think the truly "least-bad" option would be to make our own `post-type-name` block with an `isLink` attribute, and propose the same to Gutenberg. Ours can follow the same pattern as the other [blocks we've written](https://github.com/WordPress/wporg-news-2021/tree/trunk/source/wp-content/themes/wporg-news-2021/blocks), rather than a full interactive editor interface, so that shouldn't take very long to build.

StevenDufresne:
@iandunn Thanks for creating the issues in Gutenberg! 

@ryelle Yeah, I like the `post-type-name` approach. 

Maybe a `site-link` component could work? Where it outputs a link and makes use of `get_site_url? idk.

ryelle:
> Maybe a site-link component could work? Where it outputs a link and makes use of `get_site_url? idk.

While that would work conceptually in the template, it's not really a block - I don't know how that would work in an editor, for example. (edit: unless I'm misunderstanding how it would work)

StevenDufresne:
> > Maybe a site-link component could work? Where it outputs a link and makes use of `get_site_url? idk.
> 
> While that would work conceptually in the template, it's not really a block - I don't know how that would work in an editor, for example. (edit: unless I'm misunderstanding how it would work)

Yeah, it would probably break "block" conventions. I'm concerned that creating a `post-type-name` component is only kicking the can down the road, so to speak. There will be other scenarios where we want to link to things that are internal and not post types. 

Maybe thinking about this from a template first perspective is wrong? Maybe content with links should be done in the content/editor anyways. I'm behind on the localization conversation, but I imagine its a similar problem?

ryelle:
> I'm concerned that creating a post-type-name component is only kicking the can down the road, so to speak.

I see it as analogous to the `post-author-name`, `post-title`, `site-title`, etc blocks, which all have an `isLink` attribute, so I think that's an established pattern in GB.

> Maybe content with links should be done in the content/editor anyways.

This particular link isn't in the Episode CPT, it's in the template itself. We haven't been using the site editor (it's currently broken, in fact #95). If we update the template in the site editor, we're basically forking it, and that template won't get any new code updates (there's an issue in GB about this: https://github.com/WordPress/gutenberg/issues/22469). So unfortunately I don't think that's viable with the setup we have right now.

StevenDufresne:
> If we update the template in the site editor, we're basically forking it, and that template won't get any .

Ah that's right. Thanks for mentioning that? I'm not opposed to the `post-type-name` control to solve this in the short term but It would be nice and forward-moving if we can come up with something reliable. Others will be in the same position we are.

ryelle:
> It would be nice and forward-moving if we can come up with something reliable.

I don't see `post-type-name` itself as a short-term fix, I think it lines up with how other content-blocks work (I see our likely implementation as a short-term fix, so that we don't need to build a whole editor interface). I'm having trouble understanding what the other issues you see are, but if you think this is something other people would run into, you should open an issue in Gutenberg to discuss it.

StevenDufresne:
Will be replaced by #361

PR #337, opened 2022-02-17
Events category template

### LINKED ISSUE #281: Implement Events category design, posts grouped by year
Is the approach taken for `Releases` (in #246) something we can apply fairly easily to `Events`? Asking because the design is identical, with the difference that numbers on the left signal the year and group posts from that year. If not, I believe the Events category page can be addressed post launch.

Captura de ecrã 2022-02-02, às 16 22 55

coreymckrill:
Implements the design for the Events category archive. Similar to the Releases archive, except that the "version" (year in this case) is only shown once for all of the posts in the given year, instead of for each post. This presents an extra challenge, since the query block output isn't filterable, so post list items from the same year can't be grouped into a container element.

The thing I haven't been able to figure out is how to have the year block animate and highlight when _any_ post in the year group is focused, rather than just the post that contains the visible year block. (See comment below with video showing desired behavior)

Fixes #281 

To test:

* If testing in the local dev environment, you'll need to add some extra posts in the Events category to really get a sense for it. The sample data doesn't have enough.
* You can also change the Reading settings to limit the posts shown per page to see how the year groupings work when split between pages.

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/block-template-parts/content-category-events.html
+++ b/source/wp-content/themes/wporg-news-2021/block-template-parts/content-category-events.html
@@ -1,17 +1,11 @@
+<!-- wp:wporg/event-year /-->
+
 <!-- wp:group {"tagName":"header","className":"entry-header"} -->
 <header class="wp-block-group entry-header">
 	<!-- wp:post-title {"level":2,"isLink":true} /-->
 
-	<!-- wp:group {"className":"entry-meta"} -->
-	<div class="wp-block-group entry-meta">
-		<!-- wp:post-date /-->
-	</div>
-	<!-- /wp:group -->
+	<!-- wp:post-date /-->
 </header>
 <!-- /wp:group -->
 
-<!-- wp:group {"tagName":"section"} -->
-<section class="wp-block-group">
-	<!-- wp:post-excerpt {"moreText":"Read Post","showMoreOnNewLine":true,"layout":{"inherit":true}} /-->
-</section>
-<!-- /wp:group -->
+<!-- wp:post-excerpt {"moreText":"Read Post","showMoreOnNewLine":true,"layout":{"inherit":true}} /-->
--- a/source/wp-content/themes/wporg-news-2021/block-templates/category-events.html
+++ b/source/wp-content/themes/wporg-news-2021/block-templates/category-events.html
@@ -2,7 +2,7 @@
 
 <!-- wp:query {"tagName":"main","className":"site-content-container"} -->
 <main class="wp-block-query site-content-container">
-	<!-- wp:post-template -->
+	<!-- wp:post-template {"align":"wide"} -->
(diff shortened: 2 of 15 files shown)


coreymckrill:
These are the hover/focus interactions for this template prescribed in the design:
https://video.wordpress.com/v/xUventCk

I am struggling to find a good way to trigger the animation of the year block when hovering on _any_ post within a particular year, rather than just on the last post in the year that's actually displaying the year block.

beafialho:
This is looking really good @coreymckrill! Some comments:

- The page's first post should have a hover effect _(blue, underline, arrow showing up)_ as well
- Let's reduce the year's letter spacing

| Current | Mockup |
| ---- | ---- |
|  Captura de ecrã 2022-02-18, às 10 34 44   |   Captura de ecrã 2022-02-18, às 10 35 58  |

- This is a bit of a nitpick but it would be great if on animation, the year could move a little further up from the line

| Current | Mockup |
| ---- | ---- |
|   Captura de ecrã 2022-02-18, às 10 40 31  |    Captura de ecrã 2022-02-18, às 10 41 34   |

- Lastly, can you make sure the color of the year on animation is `#213FD4`

Mobile looks accurate to the mockups 👍

ryelle:
> I am struggling to find a good way to trigger the animation of the year block when hovering on any post within a particular year, rather than just on the last post in the year that's actually displaying the year block.

I tried a few different CSS selectors but I agree, without wrapping these there's no good way to add the effect to just a single year's posts.

I wonder if the best solution would be to bring in a JS snippet & put the `post-year-*` class on these too, and toggle some `active` class when any post is hover/focused? or use JS to split out the lists into per-year `ul`s, but that might be tricky style-wise.

I had one bug while testing where `beige-2` was undefined. Not sure why, it was set in `theme.json` - could be caching, the output of theme.json is cached, but I didn't dive into too deeply. I can if I see it again though.

ryelle:
Oh also, can you run `yarn workspace wporg-news-2021-theme lint:css --fix`? There are a few spacing issues my editor flagged.

coreymckrill:
@beafialho Ok, I think I've addressed all of your feedback, give it another whirl and let me know what you think!

@ryelle I tried your suggestion about the JS, and I've got something working. Would love your feedback on it.

beafialho:
@coreymckrill apologies for the inconvenience but I can't seem to get my testing env working locally 😞 Could you paste screenshots of the updates?

coreymckrill:
@beafialho No prob. I figured a screencast might illustrate the changes better, but I can add screenshots if there's anything specific you would like to see.

Screencast:
https://d.pr/v/r3ZfYx

beafialho:
> @beafialho No prob. I figured a screencast might illustrate the changes better, but I can add screenshots if there's anything specific you would like to see.

Thank you @coreymckrill, screencast is even better. The changes look great!

ryelle (approved):
Looks good, tested with a combo of keyboard, mouse, and screen reader interactions 👍🏻

PR #339, opened 2022-02-17
Single Podcast: Add meta links to header

### LINKED ISSUE #338: Single Podcast should show links under player
The design has a few other links under the player on the single podcast page.

> Download file | Play in new window | Duration: …

154443561-55d2a1d9-cd14-4a87-a2d6-de538152171e

This is missing from the current page. The player was also inserted using a group block + classname workaround. Now that we've added a few simple blocks, we could do the same here.

ryelle:
Fixes #338 — This PR switches the podcast player from using a group block that filters the render hook to using a custom block. This new block is just a proxy on top of the Seriously Simple Podcast player block, which is very similar to the render filter approach. The new block also has a `showMeta` attribute, to show the extra meta links, used on the single podcast page.

To test:

- View the podcast archive
- There should be players for each post, larger players on >768px, smaller on <=768px — no visible change from trunk
- Click into an episode
- There should be a small blue player in the header (no change to player), with meta info: download, new window, and duration (this part's new)
- Make sure the links work

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/block-templates/archive-podcast.html
+++ b/source/wp-content/themes/wporg-news-2021/block-templates/archive-podcast.html
@@ -41,10 +41,7 @@
 			</header>
 			<!-- /wp:group -->
 
-			<!-- wp:group {"className":"podcast-player"} -->
-			<div class="wp-block-group podcast-player">
-			</div>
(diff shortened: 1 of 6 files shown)


coreymckrill (approved):
👍 Looking good locally

PR #343, opened 2022-02-21
Compress SVG files using SVGO

### LINKED ISSUE #285: Compress SVG image assets
As suggested [here](https://github.com/WordPress/wporg-news-2021/issues/143#issuecomment-1029802415).

coreymckrill:
This uses [SVGO](https://github.com/svg/svgo) to reduce the size of the images folder by about 38%, from 1.2 MB to 735 KB.

This does not produce any noticeable visual change in the images. For example, the Apple Podcasts icon was reduced by about 37.5%:

| Before | After |
| --- | --- |
| before | after |

Fixes #285


iandunn (approved):
I don't see any visual difference locally either 👍🏻

PR #348, opened 2022-02-22
Remove underline offset until future date.

### LINKED ISSUE #347: Category dropdown underline "ghosting"
- The underline size is thinner for some items
 - Different category links can start getting underlines when you're not hovering / they're not focused.

https://user-images.githubusercontent.com/767313/155062880-9c936d65-f895-4422-911f-cd872a6aa475.mov

According to @StevenDufresne it's coming from:
```
a:hover, a:focus {
	-webkit-text-decoration-line: underline;
	text-decoration-line: underline; <- This one
}

in combination with:

a {
	cursor: pointer;
	text-decoration: none;
	text-decoration-thickness: 1px;
	text-underline-offset: 0.15em; <- This one
}
```

StevenDufresne:
Closes #347 

This bug seems to only affect **Chrome** and there is an open ticket:
https://bugs.chromium.org/p/chromium/issues/detail?id=1172623

This PR just removes the offset seeing that there is no ETA on when the bug will be closed.

I also don't think the line adds enough value to come up with a workaround.

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/base/_elements.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/base/_elements.scss
@@ -3,6 +3,7 @@ a {
 	cursor: pointer;
 	text-decoration: none;
 	text-decoration-thickness: 1px;
+	text-underline-offset: 0.15em;
 
 	&:hover,
 	&:focus {
--- a/source/wp-content/themes/wporg-news-2021/sass/base/_elements.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/base/_elements.scss
@@ -3,7 +3,6 @@ a {
 	cursor: pointer;
 	text-decoration: none;
 	text-decoration-thickness: 1px;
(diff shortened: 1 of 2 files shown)


ryelle:
The difference is minor on the list of categories, but looks very different on the larger links.

| Before | After |
|--------|------|
|  |  
|  |  
|  |  
|  |  

I don't see the ghosting on the larger links though. I think the actual root cause is [this line](https://github.com/WordPress/wporg-news-2021/blob/31906fa829dd849975a620436436d42bad328883/source/wp-content/themes/wporg-news-2021/sass/components/local-header/_local-header.scss#L51), `text-decoration` which is the shorthand notation, and it overrides the set `text-decoration-thickness`. If I change it to `text-decoration-line: underline`, the ghosting goes away. It might be worth checking other places where `text-decoration:` is used by itself, in case this is happening elsewhere.

StevenDufresne:
Ok cool, updated.

ryelle (approved):

PR #349, opened 2022-02-22
Increase modularization of Sass files

### LINKED ISSUE #345: Modularize category CSS
`sass/page/_categories.scss` is difficult to work with because it's so big. Each of the categories is independent of the other, so I think splitting it into multiple files in a `./categories/` folder would be easier to work with.

coreymckrill:
Split categories up into separate files. Add index files for each subfolder. Import subfolders with one line (via the index files).

These changes make it easier to avoid any one style file getting too long.

Fixes #345


iandunn (approved):
This is much nicer, thanks!

PR #352, opened 2022-02-24
Scale font sizes for Releases/Events

iandunn:
This adds some extra left/right edge padding on wide breakpoints, and scales the fonts across all breakpoints. [More screens are planned for the future](https://github.com/WordPress/wporg-news-2021/issues/39#issuecomment-1048617643), but these ones need it the most.

See #39

### Releases

Events has the same layout, so it is pretty much the same.

https://user-images.githubusercontent.com/484068/155572924-ff3a15d8-1319-42dd-baf5-f3a89d2177a4.mov

### Bottom Banner

I hid the content area to make it easier to see (otherwise the browser scrolls up and down a lot)

https://user-images.githubusercontent.com/484068/155573546-b655e158-1073-444a-9779-9715f9c0f46e.mov

### Footer

We may want to scale the global footer as well, since the difference here looks a bit odd:

Screen Shot 2022-02-24 at 9 13 59 AM

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/components/bottom-banner/_bottom-banner.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/components/bottom-banner/_bottom-banner.scss
@@ -10,15 +10,25 @@
 	margin-top: 0;
 	gap: 0;
 	background-color: var(--wp--preset--color--white);
-	line-height: 30px;
 
 	h3 {
-		font-size: 26px;
 		margin-bottom: var(--wp--custom--margin--baseline);
+		font-size: clamp(26px, 5.4vw, 36px);
+		line-height: 1.625;
+
+		@include break-medium() {
+			font-size: clamp(26px, 2.2vw, 36px);
+		}
 	}
 
 	h3 + p {
 		margin-top: 0;
+		line-height: 1.9;
+		font-size: clamp(var(--wp--preset--font-size--normal), 3.1vw, var(--wp--preset--font-size--extra-large));
+
+		@include break-medium() {
+			font-size: clamp(var(--wp--preset--font-size--normal), 1.3vw, var(--wp--preset--font-size--extra-large));
+		}
 	}
 
 	> .wp-block-group {
--- a/source/wp-content/themes/wporg-news-2021/sass/base/_layout.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/base/_layout.scss
@@ -16,6 +16,7 @@
 // styles for various sections of the site.
 // Read more here: https://sass-lang.com/documentation/style-rules/placeholder-selectors
 %two-column-grid-container {
+
 	@include break-wide() {
 		display: grid;
 		grid-template-columns: calc(var(--wp--custom--layout--content-meta-size) - 32px) auto;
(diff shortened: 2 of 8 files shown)


coreymckrill (approved):
👍 Looks like it's working as expected in local dev.

Thinking out loud here, I wonder if it would make sense to concentrate these font size and/or padding rules into scss files specifically for those properties. The advantage would be that e.g. you would only need to edit one file to adjust the side paddings across all the views of the site, and we'd avoid situations down the road where someone changes the padding in one place but not others. I guess the downside would just be that then your styles are atomized into specific properties rather than being organized by layout, which seems more logical overall. So maybe not actually a great idea, just something to ponder 🤷

iandunn:
Yeah, I went back and forth on that. I can see it making sense for some things that are interrelated, like `font-size`, `z-index`, etc. We actually had a separate file for `z-index` until [the refactor](https://github.com/WordPress/wporg-news-2021/commit/f238d6908aecc1bfedb944fbe2cff478c6011b2c).

It definitely has the drawbacks you described, though. I kinda like the idea of using `@extend`, since the styles can be centralized, but also have a reference to them in the atomic file that they're used in. We could make more things like `%scale-fonts-releases-events` in the future, to cover different layouts, and then have them all live in a central file.

I don't feel too strongly, though.

coreymckrill:
>

coreymckrill:
> I kinda like the idea of using @extend, since the styles can be centralized, but also have a reference to them in the atomic file that they're used in. We could make more things like %scale-fonts-releases-events in the future, to cover different layouts, and then have them all live in a central file.

This sounds like a good approach to me 👍

source/wp-content/themes/wporg-news-2021/sass/components/bottom-banner/_bottom-banner.scss
```diff
@@ -10,15 +10,25 @@
 	margin-top: 0;
 	gap: 0;
 	background-color: var(--wp--preset--color--white);
-	line-height: 30px;
 
 	h3 {
-		font-size: 26px;
 		margin-bottom: var(--wp--custom--margin--baseline);
+		font-size: clamp(26px, 5.4vw, 36px);
+		line-height: 1.625em; /* stylelint-disable-line declaration-property-unit-allowed-list -- This has to be dynamic and relative to the scaled font size. */
```
ryelle: Leaving it unitless will also scale it to the font size. Functionally it's the same, a line height of `1.625em` = 1.625 * this element's font size; `1.625` = 1.625 * current element font size — the only difference is if there's a child element with a different font-size, [the inherited line-height is fixed to the parent font size](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#prefer_unitless_numbers_for_line-height_values). Unlikely here, but that unexpected behavior is why unitless is preferred.

iandunn: Ah, I didn't realize that, thanks! Fixed 👍🏻

iandunn:
@beafialho , I went ahead and merged this, but please open a new issue if you see anything you'd like changed.

beafialho:
Thank you @iandunn this looks good!

> We may want to scale the global footer as well, since the difference here looks a bit odd

I agree, the footer can be scalable as well.

PR #361, opened 2022-03-11
Add Post type name block for Podcast link

StevenDufresne:
This PR introduces a `post-type-name` component that renders the post type.

## Example
```
<!-- wp:wporg/post-type-name { "tagName": "h1", "isLink": true } /-->
```

## Screenshots
| Front End | Editor |
| --- | --- |
|  |  |

### tagName
A property that allows passing in the expected HTML tag to render the post type name in.

- Type: String
- Required: No
- Default: `span`

### isLink
If this property is true, the post type will be rendered as a link

- Type: Boolean
- Required: No
- Default: `false`

Related to #330

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/blocks/post-type-name/index.php
+++ b/source/wp-content/themes/wporg-news-2021/blocks/post-type-name/index.php
@@ -48,17 +48,9 @@ function render_block( $attributes, $content, $block ) {
  */
 function register_block() {
 	register_block_type(
-		'wporg/post-type-name',
+		__DIR__ . '/block.json',
 		array(
-			'title'           => __( 'Post Type Name', 'wporg' ),
 			'render_callback' => __NAMESPACE__ . '\render_block',
-			'uses_context'    => array( 'postId' ),
-			'attributes'      => array(
-				'tagName' => array(
-					'default' => 'span',
-					'type'    => 'string',
-				),
-			),
 		)
 	);
 }
--- a/source/wp-content/themes/wporg-news-2021/blocks/post-type-name/index.php
+++ b/source/wp-content/themes/wporg-news-2021/blocks/post-type-name/index.php
@@ -21,9 +21,9 @@ function render_block( $attributes, $content, $block ) {
 		return '';
 	}
 
-	$post_id  = $block->context['postId'];
-	$post_type = get_post_type( $post_id );
-	$post_type_obj = get_post_type_object( $post_type );
+	$post_id            = $block->context['postId'];
+	$post_type          = get_post_type( $post_id );
+	$post_type_obj      = get_post_type_object( $post_type );
 	$wrapper_attributes = get_block_wrapper_attributes();
 
 	if ( isset( $attributes['isLink'] ) && $attributes['isLink'] ) {
@@ -52,13 +52,13 @@ function register_block() {
 		array(
 			'title'           => __( 'Post Type Name', 'wporg' ),
 			'render_callback' => __NAMESPACE__ . '\render_block',
-			'uses_context'    => [ 'postId' ],
-			'attributes'    => [
-				'tagName' => [
+			'uses_context'    => array( 'postId' ),
+			'attributes'      => array(
+				'tagName' => array(
 					'default' => 'span',
-					'type' => 'string',
-				],
-			],
+					'type'    => 'string',
+				),
+			),
 		)
 	);
 }
--- a/source/wp-content/themes/wporg-news-2021/blocks/post-type-name/index.php
+++ b/source/wp-content/themes/wporg-news-2021/blocks/post-type-name/index.php
@@ -1,0 +1,89 @@
+<?php
+
+namespace WordPressdotorg\Theme\News_2021\Blocks\PostTypeName;
+
+add_action( 'init', __NAMESPACE__ . '\register_block' );
+add_action( 'enqueue_block_assets', __NAMESPACE__ . '\register_block_type_js' );
+
+/**
+ * Renders the `wporg/post-type-name` block on the server.
+ *
+ * The block render the post type name. It also optionally renders it as a link.
+ *
+ * @param array    $attributes Block attributes.
+ * @param string   $content    Block default content.
+ * @param WP_Block $block      Block instance.
+ *
+ * @return string
+ */
+function render_block( $attributes, $content, $block ) {
+	if ( ! isset( $block->context['postId'] ) ) {
+		return '';
+	}
+
+	$post_id  = $block->context['postId'];
+	$post_type = get_post_type( $post_id  );
+	$post_type_obj = get_post_type_object( $post_type );
+	$wrapper_attributes = get_block_wrapper_attributes();
+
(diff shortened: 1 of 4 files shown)


source/wp-content/themes/wporg-news-2021/blocks/post-type-name/index.php
```diff
@@ -0,0 +1,89 @@
+<?php
```
iandunn, line 1: 🤔  It might be better to have this in https://github.com/WordPress/wporg-mu-plugins/, since it's not really specific to this theme.

That might be overkill, though. I don't feel strongly.

source/wp-content/themes/wporg-news-2021/blocks/post-type-name/index.php
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	$post_id  = $block->context['postId'];
+	$post_type = get_post_type( $post_id );
+	$post_type_obj = get_post_type_object( $post_type );
+	$wrapper_attributes = get_block_wrapper_attributes();
+
+	if ( isset( $attributes['isLink'] ) && $attributes['isLink'] ) {
+		return sprintf(
+			'<a %1$s href="%2$s">%3$s</a>',
```
iandunn, line 31: So when `isLink=true` it returns an anchor, and when it's `false` the function returns a `span`, is that right?

It might be better to always return the span, and then put the link inside it when `isLink=true`. That way, there's a consistent tree for CSS selectors to target.

source/wp-content/themes/wporg-news-2021/blocks/post-type-name/index.php
```diff
@@ -0,0 +1,89 @@
...
+ * Registers the `wporg/post-type-name` block on the server.
+ */
+function register_block() {
+	register_block_type(
+		'wporg/post-type-name',
+		array(
+			'title'           => __( 'Post Type Name', 'wporg' ),
+			'render_callback' => __NAMESPACE__ . '\render_block',
+			'uses_context'    => [ 'postId' ],
+			'attributes'    => [
```
iandunn: IIRC there's a WP function that will register a block based on the `block.json`, so that you don't have to duplicate all of this.

iandunn (approved):
Looks good!

I think the element tree comment would be a good addition, but not a blocker. The linter might also fix some things if you run it.

source/wp-content/themes/wporg-news-2021/blocks/post-type-name/index.php
StevenDufresne: Since you can pass in the `tagName`, styling via the tag name is not advised. Additionally, if we wanted tag consistency, as in always output a span, we would get:

```
<span><h1>Podcast</h1></span>
```
We could omit the `<span>` if the `tagName` is supplied but that may be more complicated than just not having it?

source/wp-content/themes/wporg-news-2021/blocks/post-type-name/index.php
StevenDufresne: I admittedly copied this from the other blocks in the project. I'll take a look.

StevenDufresne:
> I think the element tree comment would be a good addition, but not a blocker. The linter might also fix some things if you run it.

I did run the linter and did it again. It doesn't seem to find anything. Is there anything suspicious that you see? I am assuming you are referring to the PHP linter.

source/wp-content/themes/wporg-news-2021/blocks/post-type-name/index.php
StevenDufresne: I think I'll leave it here for now. Should we need it, move it later.

source/wp-content/themes/wporg-news-2021/blocks/post-type-name/index.php
iandunn: Ah, that makes sense, your original version seems good 👍🏻

source/wp-content/themes/wporg-news-2021/blocks/post-type-name/index.php
iandunn: https://make.wordpress.org/core/2021/06/23/block-api-enhancements-in-wordpress-5-8/

iandunn:
Yeah, PHP. I thought it wouldn't like the double spaces at https://github.com/WordPress/wporg-news-2021/pull/361/files#diff-a0ad3bed9f48dea2bef419d8715211a7cde7c1b2242faeb137ca6d5f8ca9f2eeR24 and/or would want all 4 lines to be aligned on the `=`, but no worries if not.

source/wp-content/themes/wporg-news-2021/blocks/post-type-name/index.php
iandunn: 👍🏻

StevenDufresne:
> Yeah, PHP. I thought it wouldn't like the double spaces at https://github.com/WordPress/wporg-news-2021/pull/361/files#diff-a0ad3bed9f48dea2bef419d8715211a7cde7c1b2242faeb137ca6d5f8ca9f2eeR24 and/or would want all 4 lines to be aligned on the `=`, but no worries if not.

No, you are right. Turns out this project doesn't have `phpcs.xml`. I'll create it in another PR.

StevenDufresne:
> > Yeah, PHP. I thought it wouldn't like the double spaces at https://github.com/WordPress/wporg-news-2021/pull/361/files#diff-a0ad3bed9f48dea2bef419d8715211a7cde7c1b2242faeb137ca6d5f8ca9f2eeR24 and/or would want all 4 lines to be aligned on the `=`, but no worries if not.
> 
> No, you are right. Turns out this project doesn't have `phpcs.xml`. I'll create it in another PR.

Never mind, I was ignorant of how we were importing our tools. Something is amiss though, can't get it to work on a fresh install.

```
phpcbf -p -v
Registering sniffs in the WordPress.org Meta Coding Standards standard... ERROR: Referenced sniff "PHPCompatibilityWP" does not exist
```

Could very well be user error.

iandunn:
Sounds like you might need to install https://github.com/PHPCompatibility/PHPCompatibility, but I wouldn't consider that a blocker for merge.

iandunn:
Er, it should be https://github.com/PHPCompatibility/PHPCompatibilityWP

iandunn:
...and also https://github.com/PHPCompatibility/PHPCompatibilityParagonie

source/wp-content/themes/wporg-news-2021/blocks/post-type-name/index.php
StevenDufresne: Good call, updated in [a2f1e48](https://github.com/WordPress/wporg-news-2021/pull/361/commits/a2f1e48db6732883b5ff000ae3636fc7e521ff58)

iandunn: 👍🏻  I'll update the others [so that default attributes will work](https://github.com/WordPress/gutenberg/issues/39646)

PR #362, opened 2022-03-15
Revert font scaling

### LINKED ISSUE #356: Misalignments and visual problems on posts and content
I opened up some posts randomly and found out that there's something wrong with alignments. Perhaps that's due to some on-going tasks but I thought I'd flag these out anyway because these have been addressed before and were looking good.

- [x]  In Podcast the content width is too small

Captura de ecrã 2022-02-28, às 17 37 48

- [x]  In the posts the post meta isn't aligned with "News" in the Local Nav Bar. @iandunn maybe this is happening due to the changes in [this PR](https://github.com/WordPress/wporg-news-2021/pull/352). I think if that's the case, it's better to adopt [the initial idea](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=11920%3A47029) of keeping the elements that are either left or right aligned to the edges always that way.

Captura de ecrã 2022-02-28, às 17 44 55

- [x]  The images that are set to full-width in the editor don't look full width. They should be full-width across all sizes.

Captura de ecrã 2022-02-28, às 17 46 03

- [ ]  In Events and Releases the spacings between the post titles and post date are non existent — let's revert this to what it was before.

https://user-images.githubusercontent.com/65220155/156033714-058f97ca-b5d5-4041-a11e-8deb84a27ab9.mp4

- [ ]  In both Events and Releases, the post excerpts font size is looking way too big for a 1536px screen size. These sizes should only become bigger in really large screens, from 2000px on.

- [ ]  Let's try to use [this alignment](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=11980%3A53269) only from 2000px on, I thought that was how it was going to behave and I'm not happy with how it looks now, it seems to start getting those side margins at 1040px. For before 2000px let's use [this alignment](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=4852%3A36290).

- [ ]  At 1536px the pre footer's font size is also too big. Let's apply the rules above to that section too.

Captura de ecrã 2022-02-28, às 18 14 22

iandunn:
Fixes #356 
See #39

This reverts the font scaling parts of #352, while keeping the things like `max-width` on the content, `line-height`, and miscellaneous things that brought it closer to the mockup.

It also makes the bottom banner go to a 4-column layout at 2000px. 

https://user-images.githubusercontent.com/484068/158474560-05f7d9d8-15b3-4ea4-88dd-cdd935d52020.mov


beafialho:
Thank you @iandunn, could you confirm the line height is the same as in the blog posts? It looks a bit tighter in the recording to me. Other than that, the scaling to 4 columns seems to work well!

iandunn:
You're right that It's tighter, but as far as I can tell that [matches the mockup](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=12217%3A76645). 

**Releases body text**
*  `30.4px` for mobile
*  `25.6px` for both desktop sizes
 
**Single posts body text**
* `30.4px` across all breakpoints

**Banner body text**
* `26px` for mobile
* `30.4px` for desktop
* `25.6px` for wide desktop

The titles and post meta had similar variations, but somewhat different too. It took awhile to get them all straight and I tested many times throughout, but I could have gotten something mixed up.

Is the same Figma component reused for all the copies of the light and dark banner on all pages? If not that might also explain it, since I was bouncing around between the 4 different copies for Releases and the 4 copies for Events.

beafialho:
The mockup you're linking to shows the line height for `post excerpts` which is tighter than the [posts](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=11920%3A47029) and the banner line heights. This is the mockup for reference: https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=12216%3A71385

beafialho:
In the middle of iterations I sometimes detach components but the line height is also accurate in the [main component](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=8551%3A32738). Apologies for the confusion.

iandunn:
Ah, you're right, thanks. I think maybe where I went wrong was assuming that all sections of the 4-column banner would be the same line-height, but it looks like WP Briefing should be `27.2px` while the rest are `30.4px`? 

But then on the Podcast page, it should be `30.6px` and `34.2px`, respectively?

beafialho:
> I think maybe where I went wrong was assuming that all sections of the 4-column banner would be the same line-height

You're right though, that was my mistake. All paragraphs of the banner section should be the same line-height, which is `34px` including the Podcast page. During design I set paragraphs to 190% which is 34px.

Captura de ecrã 2022-03-21, às 15 35 15

iandunn:
Ah, ok, so just a simple `line-height: 1.9` across all breakpoints 👍🏻

iandunn:
Ok, that's all deployed now 👍🏻

PR #363, opened 2022-03-17
Add styles for lists of "props"/"contributors"

### LINKED ISSUE #355: "Contributors" styles
The design specifies color and font size that depends on how long the list of contributors is:

contributors

It appears (for the [5.9 release](https://wordpress.org/news/2022/01/josephine/) anyway) that the list of contributors is just output in a single paragraph block. So perhaps these three "breakpoint" styles should just be implemented as block variations for the paragraph block?

coreymckrill:
Registers some block styles and adds related rules to the stylesheet to match the design for "Contributors" blocks. In some release posts (major releases), the `[wpcredits]` shortcode is used, which can't have the block styles applied to it, so there's [a separate PR](https://github.com/WordPress/wordpress.org/pull/69) to add the correct classes to the container that it renders. For other release posts (minor releases), the list of props is usually generated via a script and pasted into a paragraph or code block. For these, the block style can be applied manually to be able to match the design.

Note that the design calls for the props block to be full width. It seemed to make more sense to just have the post author add the additional `alignfull` class to the block, rather than trying to match the same full width styles with another class name.

Fixes #355 

## To test

1. Edit a release post in the dev environment sample data (5.7.1 is a good example) and try applying block styles to the paragraph containing the list of props.
2. You'll also need to add `alignfull` as a custom class on the paragraph.
3. It still won't quite match the design since the props are separate by commas rather than bullets (something that's addressed for the major releases in the other PR), but it'll be close.


iandunn (approved):
LGTM 👍🏻

PR #369, opened 2022-03-22
Home: Use Release Version block to avoid overflow

### LINKED ISSUE #364: Homepage font sizes too big on many breakpoints
Reported by @vijayhardaha in #356

-------------------------------

Issue on page [WordPress.org News](https://wordpress.org/news/)

Latest Release section is not optimized for width between `>780px` & `<1280px`
--

iPad Pro (Portrait) - Not OK
Dev Tools Screenshot

iPad Air (Landscape) - Not OK
iPad Device screenshot

iPad Air (Portrait) - Not OK
iPad Device screenshot

___

Overflow problem on width `<=375px`
---

There is overflowX on `>=375px` because class `.front__latest-posts-heading` used for section heading has calculated width but doesn't not `max-width` defined to handle the `overflow`

Overflow on Dev Tools
Overflow on Screen Resize

Font too big for width between `>=375px` & `<=480px`
---

We could use a bit smaller font size to avoid `text-overflow`
> Needs testing on small width iPhone(s) to see the actual output.

Font too big

_Originally posted by @vijayhardaha in https://github.com/WordPress/wporg-news-2021/issues/356#issuecomment-1070316279_

iandunn:
Fixes #364 

The mockup calls for just the version number on the homepage, but originally we included the full title because of Gutenberg limitations. We built a custom block that will extract just the version number for the Releases category, so we can use it here too.

https://user-images.githubusercontent.com/484068/159569918-29f2102e-48b4-45ba-a1fb-2b36be7ec556.mov

cc @vijayhardaha

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/block-template-parts/front-page/latest-release.html
+++ b/source/wp-content/themes/wporg-news-2021/block-template-parts/front-page/latest-release.html
@@ -7,7 +7,7 @@
 	<!-- wp:post-template -->
 		<!-- wp:group {"tagName":"header","className":"entry-header"} -->
 		<header class="wp-block-group entry-header">
-			<!-- wp:post-title {"level":3,"isLink":true} /-->
+			<!-- wp:wporg/release-version {"tagName":"h3","isLink":true} /-->
 
 			<!-- wp:group {"className":"entry-meta"} -->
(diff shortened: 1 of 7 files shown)


beafialho:
Awesome, looks great @iandunn 👍

PR #375, opened 2022-05-10
Update the color for selected text

### LINKED ISSUE #372: Highlight color for text doesn't match the color palette
When selecting copy with a post (whether a written post or the podcast), the highlight color is red. I don't know if that was an intentional choice, but it's probably worth a designer's eye to confirm.

Screen Shot 2022-04-29 at 10 45 15 AM

ryelle:
Fixes #372. This switches the text selection color from coral red to light blue. It also forces the text color to be black when highlighted (against the light blue), since otherwise white text became unreadable.

**Screenshots**

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/base/_normalize.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/base/_normalize.scss
@@ -8,13 +8,15 @@ body {
 	-webkit-font-smoothing: antialiased;
 	-moz-osx-font-smoothing: grayscale;
 
-	// Text selection background color
+	// Text selection colors.
 	::selection {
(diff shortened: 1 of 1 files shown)


beafialho:
Thanks @ryelle! I should have clarified what I meant. 

For all the coloured backgrounds my idea was to apply either the light blue or white but with opacity. I've included all examples I could think of [here](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=12675%3A48485). The top row uses `#EFF2FF` and the bottom row uses `#FFFFFF`

Let me know if that makes sense.

Captura de ecrã 2022-05-11, às 09 58 57

PR #377, opened 2022-05-19
Remove the broken author and extra text

shaunandrews:


ryelle:
@shaunandrews The broken author block should be fixed now, is this PR still valid?

shaunandrews:
Nope! Thanks!

PR #387, opened 2022-07-21
Update CSS for Gutenberg 13.7

ryelle:
- The `wp-container-*` class is no longer added to all blocks, instead we should use the `wp-block-*` class. The selector is necessary for specificity, so we can't just remove it.
- Update default link selector — This matches the gutenberg default style, which was overriding the News style due to higher specificity.

| Before | After |
|-------|------|
| before-home | after-home |
| before-podcast | after-podcast |

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/base/_elements.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/base/_elements.scss
@@ -1,5 +1,5 @@
 // Links
-a {
+a:not(.wp-element-button) {
 	cursor: pointer;
 	text-decoration: none;
 	text-decoration-thickness: 1px;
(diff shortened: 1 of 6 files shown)


andrewserong:
Thanks for fixing this up! For visibility, just linking through that this PR was required due to the change introduced in https://github.com/WordPress/gutenberg/pull/40875 that refactored how the Layout block support works. Part of the effort in that PR is to reduce the usage of container ID-based classes in favour of semantic classnames, to help address https://github.com/WordPress/gutenberg/issues/41434 and an overall direction in https://github.com/WordPress/gutenberg/issues/38719.

Targeting  the `wp-block-` classname prefix looks safe to me, as that particular classname structure is very unlikely to change.

For further changes to how the Layout block support works, we're currently tracking work in https://github.com/WordPress/gutenberg/issues/39336. One of the goals is to settle on stable classnames for layout that can be relied upon in themes, and then document these reliable classnames in the handbook. For now, it's great to see that the styling rules here can safely rely upon the `wp-block-` prefix instead!

adamwoodnz (approved):
LGTM too

iandunn (approved):
LGTM 👍🏻

PR #412, opened 2023-12-19
Updated css for Responsive Design 

pitamdey:
URL : https://wordpress.org/news/category/community/
In Responsive size the section which doesn't have any featured image and has only text is aligned end so it is showing space at the top. So if we do align the centre then it will look better
Screenshot : https://drive.google.com/file/d/1Kdcp6qUFrU6WbI-8N_8OjbS3ucv-Ybgd/view?usp=sharing


ryelle:
Thanks for the PR! However, this was the intended design, as you can see [in figma](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News?type=design&node-id=4860-45855&mode=design&t=lRh4wYUXiA7oNKXw-4), so I'm going to close this.

PR #413, opened 2024-02-08
Reduce the height of the homepage header.

### LINKED ISSUE #394: News homepage: "News" heading is too tall
When I land on News page (both on MacBook and monitor) there is content missing that should be visible on the hero viewport when user lands on this page. Can we fix it so the first view when we land on [News](https://wordpress.org/news/) is as shown below in the screenshots?

To replicate the issue, simply go to [News](https://wordpress.org/news/) page and see where the content is cut. 

**Live version (screenshot)**
live-hero-viewport

**Expected look (screenshot)**
desired-hero-viewport

**Fixing note**
If you reduce the container height in the code, ensure that the title 'News' is approximately centered between the navigation and the border/separator line below it. 

**Browsers**
 - Chrome. Lacks bottom content fully.
 - Safari. Top of the CTA and date is a bit visible. 
 
 If I should ping someone specifically for this, please let me know.

StevenDufresne:
Fixes #394.

This removes the margin from the header to bring more of the content into view. It doesn't completely match the design in the original issue. All those components are not present anymore or at least with the latest post.

| Before |  After |
|--------|--------|
|  localhost_8888_ (5) | localhost_8888_ (3) | 

## Considerations

The "News" isn't completely centered in the top container because the subnav is the same color. That matches newer designs of developer.wordpress.org and documentation. Does that work? I can remove padding from the top of the section to make it even. Looks fine to me though.

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/sass/page/front-page/_front-page.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/page/front-page/_front-page.scss
@@ -16,7 +16,6 @@ body.news-front-page {
 		z-index: 1;
 		background-color: var(--wp--preset--color--blue-1);
 		color: var(--wp--preset--color--white);
-		margin: var(--wp--custom--alignment--edge-spacing) 0 calc(var(--wp--custom--alignment--edge-spacing) * 2);
 		padding: var(--wp--custom--alignment--edge-spacing);
 		font-size: clamp(80px, 25vw, 160px);
 
--- a/source/wp-content/themes/wporg-news-2021/sass/page/front-page/_front-page.scss
+++ b/source/wp-content/themes/wporg-news-2021/sass/page/front-page/_front-page.scss
@@ -16,8 +16,13 @@ body.news-front-page {
 		z-index: 1;
 		background-color: var(--wp--preset--color--blue-1);
 		color: var(--wp--preset--color--white);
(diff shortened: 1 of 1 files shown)


adamwoodnz (commented):
On my macbook with bookmarks and the global header, I think this is cut off at an awkward height; you can't see the latest posts, only the whitespace above.

Screenshot 2024-02-12 at 6 16 41 PM

I was expecting the layout to be smart and use the window vertical height to size this header so that it would work on any aspect ratio, but that might be too much work for this. Maybe there should be a little more margin so it's closer to Marko's 'Expected look' on the ticket?

adamwoodnz:
All that said, on the [3 most common monitor sizes](https://gs.statcounter.com/screen-resolution-stats/desktop/worldwide) it looks ok

### Screenshots

| 1: 1920x1080 | 2: 1366x768 | 3: 1536x864 |
|-|-|-|
| localhost_8888_(1 1920x1080) | localhost_8888_(2 1366x768) | localhost_8888_(3 1536x864) |

StevenDufresne:
I agree, seeing only white under the top isn't ideal. I can update to not use padding but instead `viewport` units.

In terms of the example screenshot, I think we should be showing the first latest post.

StevenDufresne:
Hmm, the brush smear is on the h1. Updating to viewport units is a bigger task than I would have thought. I think I'll merge this as-is, someone can follow up with additional changes.

PR #423, opened 2024-06-07
Avoid a PHP Warning in the rest api

dd32:
```
E_WARNING: count(): Parameter must be an array or an object that implements Countable in wp-content/themes/wporg-news-2021/functions.php:162

Source: GET https://wordpress.org/news/wp-json/wp/v2/posts?per_page=1
File: wp-content/themes/wporg-news-2021/functions.php:162
```

Unsure why this has suddenly started, but the above PHP Warning is happening in the REST api fairly frequently.

Looking at the WP_Query object, it appears to be uninitialised, so this seems like a shortcut we can take to avoid it..

### DIFF
--- a/source/wp-content/themes/wporg-news-2021/functions.php
+++ b/source/wp-content/themes/wporg-news-2021/functions.php
@@ -153,7 +153,7 @@ function clarify_body_classes( $classes ) {
  */
 function specify_post_classes( $classes, $extra_classes, $post_id ) {
 	global $wp_query;
-	if ( ! is_object( $wp_query ) ) {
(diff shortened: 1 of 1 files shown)


dd32:
Screenshot 2024-06-07 at 11 58 17 AM

ryelle:
Looks like this started right when I updated GB to 18.5… yep, [in 18.3 they started passing post classes back in the API](https://github.com/WordPress/gutenberg/pull/60642). I think it's safe to shortcut like this and skip returning the custom classes, since we don't really use the editor on News anyway.

ryelle (approved):

PR #431, opened 2024-10-24
Header: Remove custom local navigation, unstick global header.

### LINKED ISSUE #409: Global header remains sticky even though the shared component has been unstuck
After https://github.com/WordPress/wporg-mu-plugins/pull/466 is merged, the global header should not be sticky anymore, and only the local nav should be.

Steps to reproduce:

1. Set your mu-plugins to https://github.com/WordPress/wporg-mu-plugins/pull/466
2. Scroll the home page
3. Expect the global header to not be sticky, but it still is

Extra work will be required to make the local nav sticky, separate from the global header. https://wordpress.org/download/releases/ has almost the same header style.

ryelle:
Fixes #409 — Unstick the global header by removing the custom code & switching to the Local Navigation Bar block, to match the other redesigned block-theme sites. The local nav block has been styled with the paint stroke mask & apply the correct category colors (e.g., [Security](https://wordpress.org/news/category/security/), [Development](https://wordpress.org/news/category/development/), [Events](https://wordpress.org/news/category/events/)).

This also replaces the content in the local navigation:

- Instead of the previous "breadcrumbs" as a the homepage link in nav item, this uses the same pattern as other sites, where the site title is always visible, and the page titles appear on scroll (L2B, see https://github.com/WordPress/wporg-mu-plugins/issues/535). On the homepage, the site title appears on scroll (L1A). There are no L3 pages.
- The "Categories" menu is now using a nav menu, which is semantically more correct, but changes the style possibilities. So now the submenu opens and has been styled to be full-width, but doesn't have the paint stroke style, and doesn't dim the rest of the page. Like other submenus, clicking anywhere else closes the menu.

Since the breadcrumbs section was updated, some pages lost the `h1`. Following the pattern of the other sites, I've added `h1`s that are visually hidden, for screen readers and SEO. This does hide the current category for non-screen-reader users, as it's no longer always visible in the local nav until you scroll.

Lastly, I've updated the version in theme.json and updated some relevant properties. This enables the "sticky" setting on blocks, and a few other font style options.

**Screenshots**

| | Before | After |
|---|---|---|
| Home | before-home-wide-1-default | home-wide-1-default |
| Home, scrolled | before-home-wide-2-scroll | home-wide-2-scroll |
| Home, menu open | before-home-wide-3-open | home-wide-3-open |
| Community | before-community-mid-1-default | community-mid-1-default |
| Community, scrolled | before-community-mid-2-scroll | community-mid-2-scroll |
| Development, menu open | before-development-mid-3-open | development-mid-3-open |
| Security | security-wide-1-default | security-wide-1-default |
| Security, mobile | before-security-small-1-default | security-small-1-default |
| Security, mobile, menu open | before-security-small-3-open | security-small-3-open |

**To test**

- View any page
- There should be a global header, then the local navigation bar
- The local nav bar should have "News" (a link to home), and "Categories"
- Clicking "Categories" should open the category menu
- Clicking anywhere else should close the menu
- Scroll down
- The global header scrolls off the page, but the local navigation bar sticks to the top of the page
- The W logo and page title appear when in "sticky" position
- All pages should have an `h1` and logical heading order


StevenDufresne (approved):
Nice, thanks for these updates!

Tested. Thinks look good. 👍

I notice we still use the brush stroke on desktop. I wonder if we can remove that. It takes away a lot of the spacing below the logo in the subnav.

Screenshot 2024-10-25 at 6 29 43 AM

fcoveram:
Seconding @StevenDufresne's suggestion. The solid blueberry looks better and feels more consistency with the rest of the site.

adamwoodnz (approved):
Haven't tested, but changes LGTM

ryelle:
Thanks folks! I've removed the brush stroke, so this fully matches the rest of the sites' local navigation bars 👍🏻

